PR #4, opened 2015-10-19
Merge location and organisation search

tyom:
Rather than having two distinct search modes use the search by organisation name as a filter to narrow down results around search location.


s-block:
:+1:

PR #11, opened 2018-10-05
LGA-304 Run FALA on a staging Kubernetes namespace

rowlando:

This pull request does the following:

- Configures the Docker image built in Circle CI build job to be pushed to AWS ECR
- Configures Kubernetes staging environment
- Deploy build to Kubernetes staging environment on successful build
- Notify Slack channel of successful deployment
- Optimise the Dockerfile
- Upgrade `node-sass`

For now, I've copied the helper Circle CI scripts from `cla_public`. The plan is to make these scripts, and others, available in a convenience Circle CI image after things have settled down a bit.

The current production instance, hosted via Template Deploy, the static assets are served from an S3 bucket. See https://github.com/ministryofjustice/fala/pull/11#issuecomment-429450104 for more info. Nginx also forwards requests received on port `80` to the python application. `nginx` will be set up in a separate pull request.

### DIFF
--- a/fala/settings/base.py
+++ b/fala/settings/base.py
@@ -13,7 +13,7 @@ sys.path.insert(0, root('apps'))
 
 DEBUG = os.environ.get('DEBUG', False)
 DEBUG = DEBUG == 'True' or DEBUG is True
-DEBUG_STATIC = False
+DEBUG_STATIC = os.environ.get('DEBUG_STATIC', False)
 
 ADMINS = ()
 
--- a/fala/settings/base.py
+++ b/fala/settings/base.py
@@ -22,8 +22,9 @@ MANAGERS = ADMINS
 DATABASES = {}
 
 ALLOWED_HOSTS = [
+    '.laa-fala-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io',
     '.fala.dsd.io',
-    '.find-legal-advice.justice.gov.uk',
+    '.find-legal-advice.justice.gov.uk'
 ]
 
 LANGUAGE_CODE = 'en-gb'
--- a/kubernetes_deploy/staging/deployment.yml
+++ b/kubernetes_deploy/staging/deployment.yml
@@ -1,0 +1,20 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: laa-fala
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      app: laa-fala-app
+  template:
+    metadata:
+      labels:
+        app: laa-fala-app
+    spec:
+      containers:
+      - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-fala:master
+        name: app
+        ports:
+        - containerPort: 80
+          name: http
--- a/kubernetes_deploy/staging/deployment.yml
+++ b/kubernetes_deploy/staging/deployment.yml
@@ -18,3 +18,6 @@ spec:
         ports:
         - containerPort: 80
           name: http
+        env:
+        - name: LAALAA_API_HOST
+          value: https://staging.laalaa.dsd.io
--- a/kubernetes_deploy/staging/deployment.yml
+++ b/kubernetes_deploy/staging/deployment.yml
@@ -21,3 +21,5 @@ spec:
         env:
         - name: LAALAA_API_HOST
           value: https://staging.laalaa.dsd.io
+        - name: DEBUG_STATIC
+          value: "True"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,33 +4,40 @@ jobs:
     docker:
       - image: docker:17.03-git
     environment:
-      DOCKER_REGISTRY: "registry.service.dsd.io"
-      DOCKER_IMAGE: "fala"
+      DSD_DOCKER_REGISTRY: "registry.service.dsd.io"
+      DSD_DOCKER_IMAGE: "fala"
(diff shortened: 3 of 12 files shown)


sldblog:
This is awesome! All great changes.

I was wondering about nginx -- does the application container as-is serve static files? If yes, we could remove the nginx related content from this pull request, merge it and have a follow-up for introducing serving static files through it.

What do you think?

rowlando:
The application container as-is doesn't serve the static files, so essentially the application doesn't work. It's not styled, none of the JS kicks in, so the maps don't work.

Because we're still merging into develop, it'll be fine to split the work into two pull requests.

sldblog:
> The application container as-is doesn't serve the static files

That's weird. Do we know why `RUN cd /home/app && ./manage.py collectstatic --noinput` is there in the Dockerfile?

rowlando:
> > The application container as-is doesn't serve the static files
> 
> That's weird. Do we know why `RUN cd /home/app && ./manage.py collectstatic --noinput` is there in the Dockerfile?

The Jenkin's deploy job uses them:

```
docker cp $ASSET_CONTAINER:/home/app/$APP/static/ ./
fab application:$APP aws:cross-account environment:$environment config:./cloudformation/$APP.yaml upload_assets:./static,/
```

This uploads the static assets to S3 - see https://github.com/ministryofjustice/fala-deploy/blob/0c04eec2eb63ca871b78efccb1cf048ac553293b/fabfile.py#L262

https://github.com/ministryofjustice/fala-deploy/blob/master/docs/pillar_example.md says this:

>assets_location This key specifies the location in nginx to proxy to the S3 bucket. If not specified, the default setting is as shown below.

and https://github.com/ministryofjustice/fala-deploy/blob/master/pillar/prod/data.sls configures the assets_location to `/static`.

Now, it is possible to serve the static assets directly from the container if we set `DEBUG_STATIC = True` in https://github.com/ministryofjustice/fala/blob/develop/fala/settings/base.py#L16 as this url will be configured in the app:

```python
if settings.DEBUG_STATIC:
    urlpatterns += [
        url(r'^static/(?P<path>.*)$', 'django.views.static.serve', {'document_root': settings.STATIC_ROOT}),
    ]
```

sldblog:
🏆 Thank you for the explanation, @rowlando!

Do you think it would make sense to enable serving static files from Python, for now, to be able to make it "work" without nginx?

kubernetes_deploy/staging/deployment.yml
```diff
@@ -0,0 +1,31 @@
...
+      - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-fala:ecr-docker-registry
+        name: app
+        env:
+        - name: DEBUG
+          value: "True"
+      - name: nginx
+        image: nginx:1.14
+        volumeMounts:
+        - name: nginx-conf
+          mountPath: /etc/nginx/conf.d
```
sldblog: It would probably be easier to use nginx if we had a `Dockerfile.nginx` file in the repository and a `docker-compose.yml` file to build/use them together. What do you think?

rowlando:
> 🏆 Thank you for the explanation, @rowlando!
> 
> Do you think it would make sense to enable serving static files from Python, for now, to be able to make it "work" without nginx?

Yeap. It won't affect the current hosting of static assets, as no requests will get through to the application. Agreed?

kubernetes_deploy/staging/deployment.yml
rowlando: Yeah, whilst we're in the development stage, definitely make sense for faster feedback.

I'll pull the nginx related code bits out of this pull requests's branch.

fala/settings/base.py
```diff
@@ -22,8 +22,10 @@
 DATABASES = {}
 
 ALLOWED_HOSTS = [
+    '.laa-fala-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io',
     '.fala.dsd.io',
     '.find-legal-advice.justice.gov.uk',
+    'localhost'
```
sldblog: Could we move `localhost` to a local setting instead, please?

kubernetes_deploy/staging/deployment.yml
```diff
@@ -0,0 +1,20 @@
...
+  selector:
+    matchLabels:
+      app: laa-fala-app
+  template:
+    metadata:
+      labels:
+        app: laa-fala-app
+    spec:
+      containers:
+      - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-fala:ecr-docker-registry
```
sldblog: Should the tag be `master`, in case someone applies it by hand?

kubernetes_deploy/staging/deployment.yml
```diff
@@ -0,0 +1,27 @@
...
+        name: app
+        ports:
+        - containerPort: 80
+          name: http
+        env:
+        - name: LAALAA_API_HOST
+          value: https://staging.laalaa.dsd.io
+        - name: DEBUG_STATIC
+          value: "True"
+        - name: DEBUG
```
sldblog: Do we need `DEBUG` too?

rowlando: No, but thought it might be helpful in our only non-prod environment.

What do you think?

sldblog: I'm unsure what effects it has and our other applications don't use it. I think it might be better to turn it on consistently once we know we need it. What do you think?

rowlando: Sure.

PR #16, opened 2018-10-26
Configure health probes for nginx container

rowlando:

This pull requests configures Kubernetes to use liveness and readiness probes to determine the health of the web app container.

>Readiness and liveness probes can be used in parallel for the same container. Using both can ensure that traffic does not reach a container that is not ready for it, and that containers are restarted when they fail.

The kubelet will send the first readiness probe 5 seconds after the container starts. The kubelet will continue to run this check every 10 seconds until its ready. Nginx starts up quickly, so maybe the 5 seconds could be reduced.

The liveness probe is configured in the same way. The probe is successful if the HTTP GET request to `/ping.json` returns a 200. Kubernetes will restart the container when it's not successful.

Below, you can see the result of stopping nginx running inside the container - the kubelet restarts the container.

screen shot 2018-10-26 at 11 33 09

None


kubernetes_deploy/staging/deployment.yml
```diff
@@ -23,6 +23,20 @@ spec:
           value: https://staging.laalaa.dsd.io
       - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-fala-nginx:master
         name: nginx
+        readinessProbe:
+          httpGet:
+            path: /ping.json
```
sldblog: `/ping.json` is served by the "webapp" container, right?

In my thinking, these containers should check that they are themselves healthy and the pod will check that all containers are healthy.

I think it's not advisable to rely on other containers for readiness and liveness. Could we change that to use a static resource instead, please?

sldblog: 💭 We could add a file that serves this purpose, or use the TCP probe instead.

rowlando: https://github.com/ministryofjustice/fala/blob/master/conf/nginx.conf#L23-L25

Nginx is configured to return a 200 for `/ping.json`.

sldblog: > Nginx is configured to return a 200 for `/ping.json`.

Oh, I see, it was added earlier, not part of this pull request.

Could we step back a bit?

- I remember suggesting using `ping.json` in the "nginx" container, but you are right, it's misleading. If someone would call `https://laa-fala-staging.apps..../ping.json`, my expectation would be to ping the Python application as it's expected that everything under `/static` is "nginx" and everything else is "webapp".  

  If we would have `/static/ping.json` the origin would be explicit. What do you think?
- > direct communication to the application isn't possible because communication happens via the uwsgi protocol

  I think _now_ I understand why `socket=:8000` was bothering me. I would always assume each container is viable on its own. If we _need_ "nginx" to try out "webapp" then we would have a tightly coupled dependency and I would not like to force that on anyone.

  Can we change this back to `http=0.0.0.0:8000`, please? I think the general expectation of any web-based component would be to work with HTTP out of the box.
- I can see now that the uwsgi `ping.json` code is unused, I missed that during reviewing #15 😢 We should not have unused code in the codebase. However, if we change the protocol back to `http`, we _could_ use it again, right?

rowlando: > If we would have /static/ping.json the origin would be explicit. What do you think?

Good idea

>I think now I understand why socket=:8000 was bothering me. I would always assume each container is viable on its own. If we need "nginx" to try out "webapp" then we would have a tightly coupled dependency and I would not like to force that on anyone.

>Can we change this back to http=0.0.0.0:8000, please? I think the general expectation of any web-based component would be to work with HTTP out of the box.

You can run the application with by following the readme, with the final command being `python3 ./manage.py runserver` or `DJANGO_SETTINGS_MODULE=fala.settings.local python3 ./manage.py runserver` to use your local settings.

uWSGI is only run when in a Docker container. This is standard for Django applications. 

https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/

So the set up for local development is:

`the web client <-> Django`

and this set up for docker-compose/Kubernetes, where web server is `nginx`:

`the web client <-> the web server <-> the socket <-> uWSGI <-> Django`

The uWSGI docs says:

>in fact it’s better to use Unix sockets than ports - there’s less overhead.

This is the set up for `cla_public`:

**nginx.conf**
`uwsgi_pass  unix:////tmp/public.sock;`

**cla_public.ini**
`socket = /tmp/public.sock`

>I can see now that the uwsgi ping.json code is unused, I missed that during reviewing #15 😢 We should not have unused code in the codebase. However, if we change the protocol back to http, we could use it again, right?

I'll remove the `ping.json` endpoint from the webapp.

rowlando:
Agreed. Let's chat next week.

PR #18, opened 2018-10-30
Configure Kubernetes production deploy

rowlando:

This pull request does the following:

- Configures Kubernetes production environment
- Allows deploys to Kubernetes production environment for `master` branch only
- Notify Slack channel of successful deployment to production

`ALLOWED_HOSTS` value is now retrieved from the environment, if available. When it's not, default values are used. The defaults are a temporary workaround until we stop using the legacy Template Deploy platform.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,6 +157,10 @@ workflows:
           type: approval
           requires:
             - staging_deploy
+          filters:
+            branches:
+              only:
+                - master
       - production_deploy:
           requires:
             - production_deploy_approval
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,5 @@
 version: 2
 
-references:
-  container_config: &container_config
-    docker:
-      - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
-        environment:
-          GITHUB_TEAM_NAME_SLUG: laa-get-access
-          APPLICATION_DEPLOY_NAME: laa-fala-webapp
-
 jobs:
   build_web:
     docker:
@@ -107,7 +99,8 @@ jobs:
             python manage.py test
 
   staging_deploy:
-    <<: *container_config
+    docker:
+      - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
     steps:
       - checkout
       - run:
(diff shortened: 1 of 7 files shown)


kubernetes_deploy/production/deployment.yml
```diff
@@ -0,0 +1,54 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: laa-fala
+spec:
+  replicas: 2
```
sldblog: Should the number of production replicas be 3, as per cla_public?
I think we can reasonably downscale memory requirements for our components.

fala/settings/base.py
```diff
@@ -22,6 +22,7 @@
 DATABASES = {}
 
 ALLOWED_HOSTS = [
+    '.laa-fala-production.apps.cloud-platform-live-0.k8s.integration.dsd.io',
     '.laa-fala-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io',
```
sldblog: `ALLOWED_HOSTS` seems to be very environment-specific. What do you think about injecting it as configuration instead? The same way we do with `LAALAA_API_HOST`?

kubernetes_deploy/production/deployment.yml
```diff
@@ -0,0 +1,54 @@
...
+            port: 8000
+          initialDelaySeconds: 5
+          timeoutSeconds: 1
+          periodSeconds: 10
+        ports:
+        - containerPort: 8000
+          name: http
+        env:
+        - name: LAALAA_API_HOST
+          value: https://staging.laalaa.dsd.io
```
sldblog: Should this be `https://prod.laalaa.dsd.io`?

.circleci/config.yml
```diff
@@ -1,13 +1,5 @@
 version: 2
 
-references:
-  container_config: &container_config
-    docker:
-      - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
-        environment:
-          GITHUB_TEAM_NAME_SLUG: laa-get-access
-          APPLICATION_DEPLOY_NAME: laa-fala-webapp
```
sldblog: I assume `GITHUB_TEAM_NAME_SLUG` and `APPLICATION_DEPLOY_NAME` have been moved to CircleCI environment variables.

Do we want to do that? In this case, it seems better to be explicit about it.

kubernetes_deploy/production/deployment.yml
rowlando: Good shout

kubernetes_deploy/production/deployment.yml
rowlando: Yes! Well-spotted

.circleci/config.yml
rowlando: They're not used anywhere that I can find so they've been removed.

fala/settings/base.py
rowlando: I'm hesitant to do this until we've migrated to Kubernetes. I don't want to figure out how to do environment variables in Template Deploy if I can avoid it. Unless you already know how?

sldblog: We could do a lookup and default to the below values `['.fala.dsd.io', '.find-legal-advice.justice.gov.uk']` if it does not exist. What do you think?

rowlando: Good idea. Will do.

PR #19, opened 2018-11-02
LGA-306: Add "Use Kubernetes" documentation and split README into separate files 

rowlando:

This pull request adds a "Using Kubernetes" section to the README. See:

https://github.com/ministryofjustice/fala/blob/docs/README.md

It also breaks the README into separate files as it was getting a bit long.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,138 +1,17 @@
 # Find a Legal Adviser
 
-## Install Python 3 with Virtualenv (OSX)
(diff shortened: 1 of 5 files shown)


sldblog:
🙋‍♂️ Needs rebase.

farrepa:
Should mention be made of the traffic split and to be mindful for any non-trivial deploys? Or will it be so short-lived that we'll keep it in collective memory?

PR #20, opened 2018-11-05
Update scripts in package.json to include build and serve scripts and…

Katy600:
… update documentation in Readme to reflect this

This pull request updates the documentation for setting up FALA. While using the document I was unable to install the 'Build assets' section using the commands 'gulp build' and 'gulp serve'. Gulp is listed in the package.json but it was unable to be installed without the correct file path. 

This could have been solved by listing the file path directly in the documentation however npm supports the "scripts" property of the package.json file, so by listing it there, npm can generate the correct file path automatically on installation. 

On further discussion with Nick, there is also an updated npm that installs npx. npx makes the process even easier for individuals using documentation. By putting $ npx my-tool into the README.md instructions it is possible to just copy-paste one command with out making any other changes. 

[Introducing npx: an npm package runner ](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,13 @@ python3 ./manage.py runserver
 
 Build assets:
 ```
-gulp build
+npm run build
 ```
 
 Or run Gulp tasks (default task is `build`):
 
 ```
-gulp serve
+npm run serve
 ```
 
 Serve task will create a proxy server on port 3000 that goes via Django’s server on port 8000 enhancing it with
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "description": "Find a Legal Adviser",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "gulp build",
+    "serve": "gulp serve"
   },
   "repository": {
     "type": "git",


package-lock.json
```diff
@@ -2758,7 +2758,7 @@
         },
         "lodash": {
           "version": "1.0.2",
-          "resolved": "https://registry.npmjs.org/lodash/-/lodash-1.0.2.tgz",
+          "resolved": "http://registry.npmjs.org/lodash/-/lodash-1.0.2.tgz",
```
sldblog: 🤔 Do we know why the scheme changed from `https` to `http`?

package-lock.json
```diff
@@ -2728,9 +2728,9 @@
       "integrity": "sha1-2cjt3h2nnRJaFRt5UzuXhnY0auU=",
       "dev": true,
       "requires": {
-        "glob": "~3.1.21",
-        "lodash": "~1.0.1",
-        "minimatch": "~0.2.11"
+        "glob": "3.1.21",
```
sldblog: 🤔 `glob` has changed from `~` version to an exact version. Do we know why this has happened?

package-lock.json
Katy600: Thanks David, I did not spot this. I realised, from talking to Nick, that I had a different version of Node installed on my computer, so therefore it updated package-lock.json file with these new dependencies . I have done a git-reset and removed these changes and now installed the correct Node on my computer. I shall make sure I check my commits more thoroughly next time before I push! :)

Katy600: @sldblog Hi David, I am just replying to your comment below about the readme using Node 8. I had to quiz Nick about that as I wasn't sure. Apparently when you install Node 8, it installs the latest minor version of node 8. So that will be the same version that package-lock.json needs.

PR #21, opened 2018-12-13
Feature/code quality

farrepa:

- Add flake8 and Black to improve and monitor code quality
- Make changes to fix raised issues
- Adds a linting step to Circle that runs flake8

Commit-wise review may be easier


farrepa:
> Do you also want to add a step in Circle to fail if the formatter causes diffs in the code?

I've been considering it, but figured we should defer it until we've got flake8 embedded in our dev team workflows 🤔 (Happening this sprint)

PR #22, opened 2019-01-07
Disable feedback page (temporarily)

sldblog:

At the moment, we are getting massive spam.
We are taking the feedback page offline until we implement proper spam filters.

The `docker push` will be retried 3 times as currently we are facing 504 errors from one of the older Docker registries and the retry is the accepted workaround until we can fully migrate to Kubernetes. (😭)

Reverts the removal of `/home/app/fala/static` folder in commit 564cc51e677a92ba32866ab0ffe3f15794fb59d4 as the folder is still required for template-deploy.

### DIFF
--- a/.circleci/tag_and_push_docker_image
+++ b/.circleci/tag_and_push_docker_image
@@ -9,7 +9,7 @@ function tag_and_push() {
   echo
   echo "Tagging and pushing $tag..."
   docker tag $built_tag $tag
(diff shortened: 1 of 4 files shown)


sldblog:
> Ticketed adding a captcha or other filter and restoring changes from 564cc51e677a92ba32866ab0ffe3f15794fb59d4?

⌛️

PR #14, opened 2018-10-23
Build and push nginx and webapp images to ECR

rowlando:
## What does this pull request do?

This pull request is the pre-cursor to getting nginx to serve static assets and front the Python web app in Kubernetes.

- `Dockerfile.nginx` builds the static assets and configures nginx.
- 'Dockerfile' still requires node because Django has dependency on jinja templates
- Python web app docker image is pushed to ECR repo `laa-get-access/laa-fala-webapp`
- nginx docker image is pushed to ECR repo `laa-get-access/laa-fala-nginx`
- Deploys nginx and webapp to Kubernetes
- Nginx has healthcheck at `/ping.json` that returns 200
- Local settings have been updated so static files are served during local development
- Use docker-compose for development of nginx<->webapp configuration (see readme)

### Docker image tagging and pushing
The `tag_and_push_docker_image` script is simplified. It must be called with 2 positional arguments:

```
./tag_and_push_docker_image <built_tag> <docker_repo>
```

Two tags are pushed for all branches. These are:

- branch name
- branch name plus short sha (sha is based $CIRCLE_SHA1)

An additional tag is pushed for a branch called `master`. This is:

- $CIRCLE_SHA1 (The SHA1 hash of the last commit of the current build.)

### uWSGI buffer size

This is increased to `buffer-size=32768` as I saw `invalid request block size: 21327` in the uswgi logs.

https://stackoverflow.com/a/26941287

>By default uWSGI allocates a very small buffer (4096 bytes) for the headers of each request. If you start receiving “invalid request block size” in your logs, it could mean you need a bigger buffer. Increase it (up to 65535) with the buffer-size option.

## Later changes

The `/conf` directory stores the `nginx.conf` and `uswgi.ini`. In the next pull request, these will be moved into the `docker` directory.

PR #23, opened 2019-01-08
Expose uwsgi service on http protocol

sldblog:

In the template-deploy world, the application runs in a Docker container but the nginx service is running on the host machine.

We don't know how to reconfigure the nginx server running on template-deploy to use the `uwsgi` protocol currently exposed, so I am taking the route of reverting the configuration to use the `http` protocol.

~~This will most likely **break** the Kubernetes configuration and it will have to be fixed.~~ Kubernetes configuration has also changed to reflect the same configuration.

## Any other thoughts

Related: #22, #14

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -25,8 +25,12 @@ http {
       }
 
       location / {
-          uwsgi_pass  webapp;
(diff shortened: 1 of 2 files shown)


sldblog:
:tophat:

PR #24, opened 2019-01-09
LGA-417 Run black on circle

farrepa:

- Runs black on circle
- Moves flake8 and black to a separate linting job. Black requires python>=3.6 while the app and tests are on python==3.4

- Reformats with Black so branch passes on Circle.
- flake8 was introduced earlier


farrepa:
Closed to renaming for consistency across repos

PR #29, opened 2019-02-19
LGA-458 - Review and copy required FALA configuration to Kubernetes

said-moj:

Added the following configuration to the kubernetes deployment:
SECRET_KEY
ZENDESK_API_USERNAME
ZENDESK_API_TOKEN

This includes changes to make the production application fail to boot without the required secrets. For production the DJANGO_SETTINGS_MODULE env var is set in the deployment.yml

### DIFF
--- a/kubernetes_deploy/production/deployment.yml
+++ b/kubernetes_deploy/production/deployment.yml
@@ -40,6 +40,26 @@ spec:
             secretKeyRef:
               name: sentry
               key: dsn
+        - name: SECRET_KEY
+          valueFrom:
+            secretKeyRef:
+              name: secret-key
+              key: SECRET_KEY
+        - name: ZENDESK_API_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: zendesk-api
+              key: ZENDESK_API_TOKEN
+        - name: ZENDESK_API_USERNAME
+          valueFrom:
+            secretKeyRef:
+              name: zendesk-api
+              key: ZENDESK_API_USERNAME
+        - name: GA_ID
+          valueFrom:
+            secretKeyRef:
+              name: ga-id
+              key: GA_ID
       - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-get-access/laa-fala-nginx:master
         name: nginx
         readinessProbe:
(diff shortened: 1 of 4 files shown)


kubernetes_deploy/production/deployment.yml
```diff
@@ -40,6 +40,26 @@ spec:
...
+        - name: ZENDESK_API_USERNAME
+          valueFrom:
+            secretKeyRef:
+              name: zendesk-api
+              key: ZENDESK_API_USERNAME
+        - name: GA_ID
+          valueFrom:
+            secretKeyRef:
+              name: ga-id
+              key: GA_ID
```
sldblog: 🙋‍♂️ The google analytics ID will be exposed on the page source, so it does not have to be a secret.

kubernetes_deploy/production/deployment.yml
```diff
@@ -40,6 +40,26 @@ spec:
             secretKeyRef:
               name: sentry
               key: dsn
+        - name: SECRET_KEY
+          valueFrom:
+            secretKeyRef:
+              name: secret-key
+              key: SECRET_KEY
```
sldblog: Is it possible to make the production application fail to boot without the required secrets?

kubernetes_deploy/production/deployment.yml
said-moj: Done

kubernetes_deploy/production/deployment.yml
said-moj: Done

Dockerfile
```diff
@@ -40,7 +40,7 @@ RUN npm install
 COPY . .
 
 RUN ./node_modules/.bin/gulp build --production && \
-    ./manage.py collectstatic --noinput
+    DJANGO_SETTINGS_MODULE=fala.settings.circle ./manage.py collectstatic --noinput
```
sldblog: Using `DJANGO_SETTINGS_MODULE=fala.settings.circle` in the `Dockerfile` we use to build a production image looks a bit weird.

I would assume we need this because the default behaviour is to "fail if things are missing"?

What do you think about flipping the relationship and make only the production environment fail for missing variables?

said-moj: > would assume we need this because the default behaviour is to "fail if things are missing"
> ?
Yes

Makes sense. I added a fala.settings.production, this is set in the DJANGO_SETTINGS_MODULE env var in the deployment.yml

PR #30, opened 2019-02-19
LGA-397 - Setup pod monitoring for Kubernetes

said-moj:
For each display

- CPU utilisation %
- memory used GiB (cached GiB as well, if it makes any sense)
- memory %
- network receive & send MiB/s
- IO read & write MiB/s (may be a stretch goal)
- 

For each metric

- also displays red lines for hard limits for these metrics
- also displays yellow lines for soft (warning) limits for these metrics

### DIFF
--- a/.circleci/deploy_to_kubernetes
+++ b/.circleci/deploy_to_kubernetes
@@ -30,4 +30,6 @@ kubectl set image --filename="$NAMESPACE_DIR/deployment.yml" --local --output=ya
   kubectl apply \
     --filename=/dev/stdin \
     --filename="$NAMESPACE_DIR/service.yml" \
-    --filename="$NAMESPACE_DIR/ingress.yml"
+    --filename="$NAMESPACE_DIR/ingress.yml" \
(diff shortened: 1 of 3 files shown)


.circleci/deploy_to_kubernetes
```diff
@@ -30,4 +30,6 @@ kubectl set image --filename="$NAMESPACE_DIR/deployment.yml" --local --output=ya
   kubectl apply \
     --filename=/dev/stdin \
     --filename="$NAMESPACE_DIR/service.yml" \
-    --filename="$NAMESPACE_DIR/ingress.yml"
+    --filename="$NAMESPACE_DIR/ingress.yml" \
+    --filename="$NAMESPACE_DIR/dashboard.yml"
```
sldblog: 🙋‍♂️I recently learned that we can also use
```yaml
kubectl apply \
  --filename="$NAMESPACE_DIR/" \
  --filename=/dev/stdin
```

PR #32, opened 2019-03-04
Enable TLS using new cloud platform spec

said-moj:

Enable TLS in the Ingress resource.

This was implicitly been done by the cloud platforms team before however recent changes mean that we need to explicitly do this

### DIFF
--- a/kubernetes_deploy/staging/certificate.yml
+++ b/kubernetes_deploy/staging/certificate.yml
@@ -1,0 +1,2 @@
+# Empty file is required as files need to exist on both staging and
+# production for the deployment
--- a/kubernetes_deploy/staging/ingress.yml
+++ b/kubernetes_deploy/staging/ingress.yml
@@ -4,6 +4,9 @@ metadata:
   name: laa-fala
   namespace: laa-fala-staging
 spec:
+  tls:
+  - hosts:
+    - laa-fala-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io
   rules:
   - host: laa-fala-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io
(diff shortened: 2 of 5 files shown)


kubernetes_deploy/staging/ingress.yml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: laa-fala
   namespace: laa-fala-staging
 spec:
+  tls:
+  - hosts:
+    - laa-fala-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io
```
sldblog: Do we need TLS for staging?

said-moj: We currently do have it on staging, this will maintain that. Also if it's free and can't see downside of having it

alkar: By default, we redirect http to https (so as to enforce https access). You'll need it for all environments I'm afraid.

sldblog: Thank you, both! I wanted to know whether the changes for staging we were required and if yes, why, but I was not clear on this intent :sweat_smile:

kubernetes_deploy/staging/certificate.yml
```diff
@@ -0,0 +1,2 @@
+# Empty file is required as files need to exist on both staging and
+# production for the deployment
```
sldblog: Public now has a way to get around this limitation :)

sldblog:
@said-moj Could this pull request description be updated with the following, please?

- the description for certificate changes (and why)
- link to the extension of circleci service account to enable the certificate changes

I think it’s better to call these out explicitly.

PR #40, opened 2019-07-02
LGA-148 - Use Google Maps API key from kubernetes

exonian:
Injects the api key into the env from kubernetes, on staging and
production
Defaults to the empty string when absent (which produces the existing
behaviour of warning dialogue and watermark when passed as the key)

Required.

Intentionally left blank.

## Checklist

### DIFF
--- a/fala/templates/base.html
+++ b/fala/templates/base.html
@@ -89,7 +89,7 @@
       ga('send', 'pageview');
     </script>
   {% endif %}
-  <script src="//maps.googleapis.com/maps/api/js"></script>
+  <script src="//maps.googleapis.com/maps/api/js?key={{ GOOGLE_MAPS_API_KEY }}"></script>
   <script>
     window.LABELS = {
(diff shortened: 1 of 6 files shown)


fala/templates/base.html
```diff
@@ -89,7 +89,7 @@ <h1>
       ga('send', 'pageview');
     </script>
   {% endif %}
-  <script src="//maps.googleapis.com/maps/api/js"></script>
+  <script src="//maps.googleapis.com/maps/api/js?key={{ GOOGLE_MAPS_API_KEY }}"></script>
```
sldblog: Since this key is publicly visible, do we need to use secrets?

Do we have domain access control setup?

exonian: We do.

And no, we don't need to use secrets. I went with them partly because they were set up initially, partly because it felt like a neat way of separating the config, and partly because I didn't really think about the fact that they are public.

PR #45, opened 2019-09-05
LGA 790 remove gateway text

joshlow:

Removes the text about the mandatory gateway as users are not required to go to CLA for education and discrimination cases

I've removed the text - do I need to remove the additional code tags around it?

## Checklist

LGA 790
https://dsdmoj.atlassian.net/browse/LGA-790

### DIFF
--- a/fala/templates/base.html
+++ b/fala/templates/base.html
@@ -65,12 +65,6 @@
     {% endblock %}
     {% block search_form %}{% endblock %}
 
-    {% if not data.count %}
-      <p>
-        For advice on education or discrimination contact
-        <a href="https://www.gov.uk/civil-legal-advice" rel="external" target="_blank">Civil Legal Advice.</a>
-      </p>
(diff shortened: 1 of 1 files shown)


fala/templates/base.html
```diff
@@ -66,10 +66,7 @@ <h1>
     {% block search_form %}{% endblock %}
 
     {% if not data.count %}
-      <p>
-        For advice on education or discrimination contact
-        <a href="https://www.gov.uk/civil-legal-advice" rel="external" target="_blank">Civil Legal Advice.</a>
-      </p>
+
```
exonian: Yep, as you wondered, there's no point in the `{% if ... %}` and `{% endif %}` statements now, so they should be removed too.

joshlow: Thanks @exonian - i've removed the if statements. Can you look again?

PR #49, opened 2019-11-08
LGA-699 - Migrate staging to live-1 environment

exonian:

- Adds a live_1 deployment pathway within the circleci workflow
- Replaces the `deploy_to_kubernetes` script and moves the existing one so that is still be available to live_0 deploys
- Moves the staging k8s config for live_0 deployment so that the `staging` config refers to live_1
- Uses a new `laa-fala-live-1` `Context` created on the Circle CI site, to manage env vars
- Adds local k8s configuration
- Removes the nginx container from our deployment and replaces it with usage of an s3 backend
- Adds a `Job` to manage static files at the start of a deployment
- Processes static file urls in css and js files to replace them with backend-aware ones (necessary because our buckets are forced to be private-read, so we need signed urls)
- Adds a cronjob to run the static file management daily, because the longest we can sign urls for is a week 

Intentionally left blank.

## Checklist


fala/settings/base.py
```diff
@@ -99,7 +109,7 @@ def root(*x):
 
 INSTALLED_APPS = ("django.contrib.sessions", "django.contrib.messages", "django.contrib.staticfiles", "requests")
 
-PROJECT_APPS = ("adviser", "laalaa")
+PROJECT_APPS = ("adviser", "fala", "laalaa")
```
said-moj: Is this adding a new fala app? I can't seem to find a fala app inside the `apps` directory?

.circleci/config.yml
```diff
@@ -67,90 +134,105 @@ jobs:
...
-  staging_deploy:
+  live-1_staging_deploy:
     docker:
       - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
     steps:
       - checkout
       - run:
           name: Initialise Kubernetes staging context
           command: |
+            unset KUBE_ENV_PRODUCTION_CACERT
```
said-moj: Where do these `KUBE_ENV_PRODUCTION_`* env vars come from

fala/settings/base.py
exonian: It's adding the main app of the project itself (ie the `fala` dir which `apps` is inside)

.circleci/config.yml
exonian: These are the ones set on the circle ci site, at the moment on the project settings page, for live-0.
In the new live-1 staging flow, we override the project settings env vars with the ones from the `context`, but the context doesn't have the `_PRODUCTION_` vars in yet, just the `_STAGING_` ones.

The cloud platform tools image finds _all_ keys that start with `KUBE_ENV_`, so it was finding the live-0 production ones and trying to auth on live-1 with them

Once we've added the live-1 `_PRODUCTION_` ones to the context, we won't need to unset these. Having all of these in contexts will mean stuff like this doesn't happen in future, but for now we have to deal with the project settings env vars being fallbacks.

.circleci/config.yml
```diff
@@ -67,90 +134,105 @@ jobs:
...
+      DSD_DOCKER_REGISTRY: "registry.service.dsd.io"
+      DSD_DOCKER_IMAGE: "fala"
     steps:
       - checkout
+      - setup_remote_docker:
+          version: 17.03.0-ce
+          docker_layer_caching: true
       - run:
-          name: Setup Python environment
+          name: Login to the DSD Docker registry
```
sldblog: _not a blocker_

🙋‍♂ Probably not for now, but we can get rid of all DSD (template-deploy) remnants. It also means we can upgrade Docker to 18.09. 🎉

.circleci/config.yml
```diff
@@ -163,40 +245,66 @@ jobs:
...
+      - live-1_staging_deploy_approval:
+          type: approval
+          requires:
+            - live-1_build
+      - live-1_staging_deploy:
           requires:
-            - production_deploy_approval
+            - live-1_staging_deploy_approval
+          context: laa-fala-live-1
+      # - live-1_production_deploy_approval:
```
sldblog: It looks like the `live-1_production_deploy` job would not work just now.
What do you think of removing all related code until it works?

fala/management/commands/rewrite_paths_in_static_files.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Command(BaseCommand):
+    help = "Rewrites static file paths within our compiled css file, to be aware of our storage backend"
+    css_dir = root("assets", "css")
+    js_dir = root("assets", "scripts")
+
+    regex_patterns = {
+        "css": r"url\((?P<url>\"?(?:{static_url}|\.\.)(?P<path>.*?)\"?)\)",
+        "js": r"""url\: ?['"](?P<url>(?:{static_url})(?P<path>.*?))['"]""",
```
sldblog: _not a blocker_

🙋‍♂ It may be worth to invest in a webpack prototype spike; it has plugins which do the static rewrite in CSS and JS. This guide looks like it would be useful: https://github.com/petehunt/webpack-howto#5-stylesheets-and-images.

fala/assets-src/scripts/find-legal-adviser.js
```diff
@@ -253,9 +253,9 @@ exports.FindLegalAdviser = {
     var image;
     options = options || {};
 
-    if(options.icon) {
+    if(options.icon && options.icon==='icon-location-2x') {
```
sldblog: _not a blocker_

I wonder if there's any circumstance we call _without_ `icon`? "icon" may be an unnecessary configuration and we could default to `icon-location-2x`, unless `id` is given.

Dockerfile
```diff
@@ -43,5 +46,9 @@ COPY . .
 RUN ./node_modules/.bin/gulp build --production && \
     ./manage.py collectstatic --noinput
 
+# Project permissions
+RUN  chown -R app: /home/app
```
sldblog: _not a blocker_

Does the application write to any of the folders?

If not, we don't need to do this. The downside of doing a wide-sweep `chown` is that it essentially adds the size of the `/home/app` again as a new layer, increasing the overall image size.

.circleci/deploy_to_kubernetes_live_0
```diff
@@ -0,0 +1,36 @@
...
+
+ROOT=$(dirname "$0")
+NAMESPACE="$1"
+NAMESPACE_DIR="$ROOT/../kubernetes_deploy/$NAMESPACE"
+WEBAPP_REPO="$2"
+NGINX_REPO="$3"
+
+if ! [ $NAMESPACE ] ; then
+
+  echo "usage: deploy_to_kubernetes namespace\n"
```
sldblog: _nitpick_ Inconsistent filename due to the rename. `$0` may be safer.

_newlines_
`echo "text\n"` does not behave the same way as on a Mac, it will need an extra `echo -e` argument for the newline to work:

```
$ docker run -ti ubuntu:trusty-20180712 echo "test\n"
test\n

$ docker run -ti ubuntu:trusty-20180712 echo -e "test\n"
test

```

🙋‍♂ `shellcheck` is really good to pick out these:

```
$ shellcheck demo.sh

In demo.sh line 3:
echo "asd\n"
     ^-----^ SC2028: echo may not expand escape sequences. Use printf.

For more information:
  https://www.shellcheck.net/wiki/SC2028 -- echo may not expand escape sequen...
```

.circleci/config.yml
sldblog: It sounds like it'd be worth setting up a `context: laa-fala-live-0` to allow composition instead of exclusion. Is that possible to do?

exonian: It's possible but for it to have any positive effect on the live-1 deploy, we'd have to move the live-0 secrets from the repo to the context. We would also have to modify the live-0 deploy steps to use the context, which is simple enough. It doesn't seem worthwhile for the timescales involved

To clarify, we only needed to exclude them while the live-1 context didn't have keys to override them. It now does, so the unsetting is redundant (but harmless) now. I think I'll remove it, but leave the keys where they are (and only have a context for live-1).

In general, yes, composition is definitely the way to go.

.circleci/config.yml
exonian: Not particularly bothered either way

.circleci/deploy_to_kubernetes_live_0
exonian: Good spots. These are more relevant to the other file, as this is the renamed existing one which will be going away soon. I'll bear them in mind for future work

fala/management/commands/rewrite_paths_in_static_files.py
exonian: Yes that's a very good idea, hadn't realised it had the image-encoding feature. We would have to make sure we could set that limit higher than anything we'd encounter, as we'd need it to always do that and never use `url()`

PR #56, opened 2020-01-21
LGA-943 - Upgrade to Django 2.2

said-moj:

Upgrade to Django 2.2
Update python from 3.4 to 3.7 (django 2.2 requires at least 3.5)

Intentionally left blank.

## Checklist

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,5 +1,5 @@
 boto3==1.10.8
-django~=2.2
+django==2.2.10
 django-storages==1.6.5
 raven==6.9.0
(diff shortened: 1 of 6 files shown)


requirements/base.txt
```diff
@@ -1,5 +1,5 @@
 boto3==1.10.8
-django==1.11.23
+django~=2.2
```
exonian: I would prefer to pin to exact version. I like looser pinning when there's a manifest and a lock, but when there's only one file I think it's best to be exact.

PR #59, opened 2020-03-12
LGA-1032 header and footer

MalcolmVonMoJ:

Updates header and footer to new style

Moved the remote templates generic-py-base.html  and external.html from the "mojular-templates" nodejs package into the project so they can be edited.

Removed grey background from form area.

## Checklist

### DIFF
--- a/fala/templates/external.html
+++ b/fala/templates/external.html
@@ -1,0 +1,29 @@
+{% extends 'generic-py-base.html' %}
+
+{% block head %}
+  {{ super() }}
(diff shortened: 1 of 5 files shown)


fala/templates/external.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'generic-py-base.html' %}
+
```
said-moj: Is the removal of {% load staticfiles %} deliberate here?

PR #71, opened 2020-06-05
LGA-1191 error styling

MalcolmVonMoJ:

- Added error highlighting for input boxes.
- Added error text for global errors (visible above postcode input and screen reader only above org input)
- Restricted red line for non-global errors to relevant field only.
- Error summary for no answer changed to link to both postcode and org boxes.

Intentionally left blank.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,10 +155,9 @@ workflows:
           context: laa-fala-live-1
       - staging_deploy_approval:
           type: approval
-          requires:
-            - build
       - staging_deploy:
           requires:
+            - build
             - staging_deploy_approval
           context: laa-fala-live-1
       - production_deploy_approval:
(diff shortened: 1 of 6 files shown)


itzcookiie:
There is no error message appearing above the 'organisation name field' when no value is entered. Does it need to be precisely as shown in the picture on the JIRA board?

image

MalcolmVonMoJ:
After feedback from Beverley, we have moved the error message in this particular case to be above both questions - so it only appears once and refers to both, but doesn't put an emphasis on one over the other.  
image

mayowaadeniyi:
Uh I ended up approving kinda early again. Sorry Malcolm... but I wanted to ask is there a particular reason why we have "Enter a postcode or an organisation name" shown in two places?

PR #80, opened 2020-09-21
LGA-1426 - Sentry SDK

exonian:
Switches from Raven library to Sentry SDK, and starts passing an `environment` tag.

Uses the same secret name for DSN, so merging this PR will still send data to the old Sentry instance, eg https://sentry.service.dsd.io/mojds/staging-fala/issues/63232/ (triggered via a deliberate exception-causing path that was temporarily added to the view code)
Once this is merged, we can reroute errors to the new Sentry instance by updating the secret.
The new instance has been tested on staging by using a different secret name in order to route to it before the existing secret is changed eg https://sentry.io/organizations/ministryofjustice/issues/1901244840/?project=5433781&query=is%3Aunresolved

## Checklist

### DIFF
--- a/fala/apps/adviser/views.py
+++ b/fala/apps/adviser/views.py
@@ -15,9 +15,6 @@ class AdviserView(TemplateView):
         view_name = resolve(request.path_info).url_name
         current_url = reverse(view_name)
 
-        if "error" in request.GET.keys():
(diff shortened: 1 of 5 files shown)


said-moj (approved):
👍 LGTM

PR #83, opened 2020-10-08
Feature/LGA-1395 Environment Badge

mayowaadeniyi:

- Adds a graphical indicator and badge to display the current working environment 
- Changes the name of the badge on production (and eventually live) from **Alpha** to **Beta**

The name of the badge on production (and eventually live) has been changed to **Beta** from **Alpha** due to changes discussed outside of the ticket.

**You will have to recreate your `local.py` file from `local.example.py` in order for these changes to take fully effect.** This is because we do not want the default environment to be set to dev if for example, we was on a production environment and the production env var was removed, we would be seeing a badge of dev on production. 

Not creating your `local.py` will cause your badge to be **Unknown** instead of **Development**

## Checklist

### DIFF
--- a/fala/settings/base.py
+++ b/fala/settings/base.py
@@ -142,7 +142,7 @@ if "SENTRY_DSN" in os.environ:
 
 LAALAA_API_HOST = os.environ.get("LAALAA_API_HOST", None)
 
-CLA_ENV = os.environ.get("CLA_ENV", "dev")
+ENVIRONMENT = os.environ.get("ENVIRONMENT", "dev")
 
 # Zendesk settings for feedback
 ZENDESK_API_USERNAME = os.environ.get("ZENDESK_API_USERNAME")
--- a/fala/settings/base.py
+++ b/fala/settings/base.py
@@ -142,7 +142,7 @@ if "SENTRY_DSN" in os.environ:
 
 LAALAA_API_HOST = os.environ.get("LAALAA_API_HOST", None)
 
-ENVIRONMENT = os.environ.get("ENVIRONMENT", "dev")
+ENVIRONMENT = os.environ.get("ENVIRONMENT", "unknown")
 
 # Zendesk settings for feedback
 ZENDESK_API_USERNAME = os.environ.get("ZENDESK_API_USERNAME")
--- a/fala/settings/base.py
+++ b/fala/settings/base.py
@@ -142,6 +142,8 @@ if "SENTRY_DSN" in os.environ:
 
 LAALAA_API_HOST = os.environ.get("LAALAA_API_HOST", None)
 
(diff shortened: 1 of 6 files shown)


fala/settings/base.py
```diff
@@ -142,6 +142,8 @@ def root(*x):
 
 LAALAA_API_HOST = os.environ.get("LAALAA_API_HOST", None)
 
+ENVIRONMENT = os.environ.get("ENVIRONMENT", "dev")
```
exonian, line 145: I would prefer to change this to "unknown" or something. It makes me nervous to assume dev, especially as dev is going to be the environment we're least cautious if we think we're in

exonian (changes requested):

fala/templates/generic-py-base.html
```diff
@@ -8,7 +8,7 @@
     {% block stylesheets %}{% endblock %}
     {% block head %}{% endblock %}
   </head>
-  <body class="govuk-template__body ">
+  <body class="govuk-template__body {{ 'cla-' + ENVIRONMENT }}">
```
MalcolmVonMoJ, line 11: I don't think we should use the prefix `cla-`, perhaps `fala-` is more suited.  The CSS will need changing too.

MalcolmVonMoJ (changes requested):

fala/settings/base.py
mayowaadeniyi: If this is changed to unknown, how we will set up an environment for dev? Should I assume an unknown environment is equal to development? Other reasoning being this is the same set up as on cla_public

exonian: > Should I assume an unknown environment is equal to development?

No, that's exactly the thing I said we shouldn't do.
If this is how it's been done on cla_public it needs fixing on cla_public

If we're going to mark an environment as "dev" there has to be _something_ explicitly stating that (ie setting the env var to that value) and not just us identifying the lack of it being set. As a thought experience: if I inadvertently delete the line from our kubernetes setup that sets the env var to "production" I *do not* want it to start saying "dev" on the site: 1) it breaks the intended experience, 2) it masks the problem that we unexpectedly don't know which environment it is

As to how to solve it, for one idea (may or may not be the best one) have a think back to this morning and how we have a `local_settings.py` file (which we were looking at an API_HOSTNAME setting in, but can be used for any setting).

exonian (approved):

fala/settings/base.py
exonian, line 145: I don't know if it was me or you, but this got reverted to "dev", I think you had it as "unknown" earlier, which is correct

exonian (changes requested):

MalcolmVonMoJ (approved):
LGTM 👍

exonian (approved):
Great stuff

PR #106, opened 2021-10-29
LGA-1897 - Add external DNS annotations to FALA ingress

said-moj:

- Add external DNS annotations to FALA ingress.
- Update circleci python image used for testing from circleci/python:3.7 to cimg/python:3.7 otherwise it was failing to build uwsgi
- Fix base image to debian buster
  - There has been a new release of debian(bullseye) since the creation of the Dockerfile
  - This release is incompatible with node8 that is built for this project
  - LGA-1909 has been created to capture the need to upgrade node for this project

Cloud Platforms is migrating from a self-hosted Kubernetes cluster to EKS.
As part of this migration, the load balancer will need to stop directing traffic to the existing instance of our service and start directing it to an instance of our service running on the new cluster. This cutover will be controlled by us, by annotations on our Ingress objects.

In preparation for this, we need to add annotations now that specify to direct all traffic to the current instance, so that when the load balancer starts reading these annotations it knows to continue as it is now until those annotations change

The `blue` suffix means to use the ingress in “live-1”

https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/migrate-to-live.html#step-2-amend-and-apply-your-quot-live-1-quot-ingress-resource

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
 
   test:
     docker:
-      - image: circleci/python:3.7
+      - image: cimg/python:3.7
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


Pheonnexx (approved):
LGTM 👍

PR #118, opened 2022-02-15
LGA-1940-Update Ingress

jogilder:
Updated version in ingress, still has references to live1 so unlikely to pass tests. Also are references to live1 under kubernetes_deploy folders

Required.

Intentionally left blank.

## Checklist


said-moj:
Sorry I think I forgot to remove the staging-live1 and production-live1 folders. These can now be removed

jogilder (commented):
OK, My plan would also be to remove lines 11 and 12 and 17 and 18 in here and the equivalent in staging.yml and the same files under kubernetes_deploy. I will do that and recommit.

kubernetes_deploy/production/deployment.yml
```diff
@@ -37,7 +37,7 @@ spec:
           name: http
         env:
         - name: ALLOWED_HOSTS
-          value: ".laa-fala-production.apps.live-1.cloud-platform.service.justice.gov.uk .find-legal-advice.justice.gov.uk"
+          value: ".laa-fala-production.apps.live.cloud-platform.service.justice.gov.uk .find-legal-advice.justice.gov.uk"
```
said-moj, line 40: I think this needs to stay as is as those live-1 domains still work on the new cluster

jogilder: ok, does it also need to reference live or just live-1? And what about in the ingress files?

PR #130, opened 2022-09-01
LGA-2126 - Update ingress to v1.22 and add modsec in detection only mode

said-moj:

- Add modsec in detection only mode
- Create new ingress that has spec.ingressClassName set to 'modsec' so that the new v122 ingress controller is used
- Pin jinja2 to 3.0.3 as new 3.1.x release breaks backward compatibility
- Pin version click to 8.0.4 supported by our black version package
- Added ./npm_install_wrapper.sh wrapper script to help with npm install

The ingress `laa-fala` on both staging and production has had it's annotation `external-dns.alpha.kubernetes.io/aws-weight`  set to 0

When this PR is merged a new `laa-fala-v122` will be created with `external-dns.alpha.kubernetes.io/aws-weight` of 100

The new ingress will take a while to setup(DNS propagation), in that time the old ingress will continue to service request.
Once the new ingress is ready, the old ingress needs to be manually removed

More information can be found on the user guide https://runbooks.cloud-platform.service.justice.gov.uk/Switch-ingress-to-v1-ingress-controller.html#switch-ingress-to-the-new-nginx-ingress-controller-v1-2

Added ./npm_install_wrapper.sh wrapper script to help with npm install because we cannot directly call npm install as some packages have not update how they bring in their dependencies as the 
unauthenticated git:// no longer works see https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git for more details

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,8 +33,9 @@ RUN pip3 install --user --requirement ./requirements.txt
 
 # Install npm dependencies
 COPY package.json package-lock.json ./
+COPY npm_install_wrapper.sh npm_install_wrapper.sh ./
 USER 1000
-RUN npm install
+RUN ./npm_install_wrapper.sh
 USER root
 
 COPY . .
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,7 +21,9 @@ virtualenv -p python3 env
 source env/bin/activate
 
 pip install -r requirements/base.txt
-npm install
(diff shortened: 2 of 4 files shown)


psweeting1 (commented):
LGTM but I think and second opinion would be good as well.

psweeting1 (approved):
LGTM

Pheonnexx (approved):

PR #163, opened 2023-04-11
LGA-2422 - ITHC update FALA containers

jogilder:
- Update requirements files so can have generated requirements files so can more easily pin any future changes

- Use [cloud-platform tools image ](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/deploying-an-app/using-circleci-for-continuous-deployment.html#building-and-pushing-image-to-ecr) to build and tag docker image (rather than use a standard docker image) as well as to deploy the image. 

- Removed installation of pip as it is already installed in docker container and assume second install was just to pin version. Would like an opinion on whether this is a problem
- sonar warnings around copy . . - should be ok as only copy in required node files to final build and using .dockerignore
- why is flake8 set to >=3.6? Can this not just be set to a particular value?

Required.

- now install aws cli v2 and this leads to a credentials warning in circleci. CP have been notified and they are happy for us to proceed and will be in touch once they have a fix for the warning.

Intentionally left blank.

## Checklist


Dockerfile
```diff
@@ -1,47 +1,42 @@
-FROM python:3.7-buster
+FROM node:8 as node_build
+
+COPY package.json package-lock.json ./
+COPY npm_install_wrapper.sh npm_install_wrapper.sh ./
```
psweeting1, line 4: This is doing two copies of the same thing.

Dockerfile
```diff
@@ -1,47 +1,42 @@
...
+RUN ./npm_install_wrapper.sh
+
+COPY . .
```
psweeting1, line 7: This should be fine, just wondering if there is a way to be more explicit in its path, Sonar may be okay with it.

psweeting1 (commented):
The flake8 version of 6.0.0 is fine with me. I think the only issue I see, if flake8 in the project is set to a low version it may miss when the higher version of flake8 can see in circle etc.

Dockerfile
jogilder: Sonar isn't ok with it - I figured it was ok as we then do the copy later on of just the assets but would like to make this more explicit if we can.

psweeting1: After have a discussion around this in person on a call, this should be fine.

psweeting1 (approved):
LGTM

sarahyjja (approved):
LGTM

PR #191, opened 2023-11-27
LGA-2757 Remove GA3 and add GTM

digitalali-moj:

Remove GA3, Use Google Tag Manager to call GA4.

Hardcoded ID as simpler and won't change. Using existing cookie framework to implement.
Using inline script as per existing approach, but this may be something of a refactor at some point to put all scripts in files.

## Checklist

### DIFF
--- a/fala/templates/feedback/index.html
+++ b/fala/templates/feedback/index.html
@@ -1,5 +1,20 @@
 {% extends 'base.html' %}
 
+{% block before_main_js %}{% endblock %}
+  <script>
+  function tagMan() {  // Activate Google Tag Manager
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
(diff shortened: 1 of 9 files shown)


fala/templates/feedback/index.html
```diff
@@ -27,11 +30,3 @@ <h1 class="govuk-heading-xl">Your feedback</h1>
   </form>
 {% endblock %}
 
-{% block after_main_js %}
-  {{ super() }}
-  {% if GA_ID and error %}
-    <script>
-      ga('send', 'event', 'error', 'feedback-error');
```
digitalali-moj, line 34: Note: This doesn't seem to be running, or maybe the whole page is not running as I couldn't find the feedback form. But I checked GA and the event wasn't pulled in anyway, so removed.

psweeting1 (approved):
LGTM

BenMillar-CGI (approved):
LGTM! 👍

PR #201, opened 2024-03-19
EL-1410 virtualenv readme, upgrade to Node 10 (as 8 didnt work) and Python 3.12

starswan:

Update virtualenv documentation. Upgrade node to version 10, as version 8 doesn't work at all

## Checklist


docs/virtual-env.md
```diff
@@ -4,7 +4,7 @@
 
 - [Virtualenv](http://www.virtualenv.org/en/latest/)
```
davidread, line 5: no longer needed

docs/virtual-env.md
```diff
@@ -4,7 +4,7 @@
...
 - [Python 3](http://www.python.org/) (Can be installed using `brew install python3`)
```
davidread, line 6: 3.10 is the latest supported by this version of uwsgi, I believe. And it's what's in the dockerfile, so we should put something in about using a supported version, perhaps further on.

Also, brew is perhaps not the best choice if we want people to be flexible on version and have different versions for different repos. Suggest we recommend pyenv - could crib from: https://github.com/ministryofjustice/cla_backend/blob/master/VIRTUAL_ENV.md#pyenv-python2 and could add a `.python-version` file

fala/settings/local.example.py
```diff
@@ -4,5 +4,7 @@
 DEBUG = True
 ALLOWED_HOSTS = ["localhost"]
 DEBUG_STATIC = True
-LAALAA_API_HOST = os.environ.get("LAALAA_API_HOST", "http://0.0.0.0:8001")
+LAALAA_API_HOST = os.environ.get(
+    "LAALAA_API_HOST", "https://laa-legal-adviser-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk"
```
davidread, line 8: nice - makes sense to point at a working one by default

docs/virtual-env.md
```diff
@@ -4,7 +4,7 @@
...
-- [nodejs.org](http://nodejs.org/) (v8.12 - can be installed using [nvm](https://github.com/creationix/nvm))
+- [nodejs.org](http://nodejs.org/) (v10.24 - can be installed using [nvm](https://github.com/creationix/nvm))
 - [docker](https://www.docker.com/) - Only required for running application from Docker
```
davidread, line 8: this isn't relevant in this file

davidread (dismissed):

Dockerfile
```diff
@@ -1,8 +1,12 @@
 FROM node:10 as node_build
 
 COPY package.json package-lock.json ./
-COPY npm_install_wrapper.sh  ./
-RUN ./npm_install_wrapper.sh
+# On the Git protocol side, unencrypted git:// offers no integrity or authentication, making it subject to tampering.
+# We expect very few people are still using this protocol, especially given that you can’t push (it’s read-only on GitHub).
+# We’ll be disabling support for this protocol.
+# https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git
+RUN git config --global url."https://".insteadOf git://
```
davidread, line 8: What's the justification for keeping this? I don't think there is a threat worth worrying about, to start monkeying with devs' global git config. This old protocol is not used in the circleci build or README for devs, so surely it just adds complication?

docs/virtual-env.md
davidread: It would be helpful to remove 'docker' while in here

docs/virtual-env.md
```diff
@@ -2,9 +2,8 @@
 
 ## Dependencies
 
-- [Virtualenv](http://www.virtualenv.org/en/latest/)
 - [Python 3](http://www.python.org/) (Can be installed using `brew install python3`)
-- [nodejs.org](http://nodejs.org/) (v8.12 - can be installed using [nvm](https://github.com/creationix/nvm))
+- [nodejs.org](http://nodejs.org/) (v10.24 - can be installed using [nvm](https://github.com/creationix/nvm))
```
davidread, line 6: How about we make this easier for people to do this with nvm, by being more explicit? e.g. https://github.com/ministryofjustice/cla_frontend/blob/master/README.md#nodejs-v89x

Dockerfile
starswan: The docker image doesn't build on circle without this - its just the Docker image, so it doesn't affect anyone's machines any more.

davidread: Ah, I tried it: https://app.circleci.com/pipelines/github/ministryofjustice/fala/1317/workflows/de9b4bd4-93c4-4707-862b-a1e96c807adc/jobs/5429/parallel-runs/0/steps/0-107 and it times out during `npm install`

So the reason isn't (just) security, it's some install issue. I guess one of the old node packages tries to get something from github using `git://`. If this is the case, then wouldn't it happen on our local machines too? Have you tested deleting your node_packages and rerunning npm install?

davidread (approved):

PR #202, opened 2024-03-19
EL-1411 Removed docker.md file and merged into main README.md with de…

psweeting1:

Running through the Fala Docker install and update & extend the documentation if required.

There was no test command provided in the README.md or anywhere. I've provided my own. 

## Checklist

[EL-1411](https://dsdmoj.atlassian.net/browse/EL-1411)


README.md
```diff
@@ -10,8 +10,64 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+
+Please ensure your internet network such as your VPN does not interfere with your build.
+
+To run container:
+
+`./run_local.sh`
+
+This shell script contains commands that build and runs the fala app.
+
+If you're not using `Docker Desktop`, you may want to have two terminals open for the following:
```
davidread, line 36: Oh, I wasn't aware Docker Desktop was an option.

Rather than being unopinionated, I think it's good if Readme recommends a particular way, that we we know works well, can improve together, and swap tips on. So I'd encourage you to add a recommendation on one of the two ways

README.md
```diff
@@ -10,8 +10,64 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+
+    If you added pdb breakpoints, this is where you can interact with this debugger.
+
+Alternatively, some editors have functionality to hook into running containers, such as VS Code's 'Dev Containers' 
+extension. Docker Desktop offers a UI version of looking through logs, exec and many other interactions. 
+
+### Debugging
+Depending on how you're running the project, via `venv` or Docker you can perform the following debugging:
+
+Add import pdb; pdb.set_trace() as a 'breakpoint' line in the code, where you want to trigger the debugger.
```
davidread, line 60: The docs say either works https://docs.python.org/3/library/pdb.html, but in Python3 now, it's generally preferred to use `breakpoint()`, instead of `import pdb; pdb.set_trace()`

davidread: Also, for debugging to work in docker, you need to add this:
```
diff --git a/docker-compose.yml b/docker-compose.yml
index feaaa39..9a0b816 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,3 +17,4 @@ services:
       DJANGO_SETTINGS_MODULE: "fala.settings.local"
     ports:
       - "8013:8000"
+    stdin_open: true
```

README.md
```diff
@@ -10,8 +10,64 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+
+Browsing the app - Point your local browser at http://localhost:8013/
+
+Log output - watch the output generated by the running app using: `docker logs fala -f` or `docker attach fala`
+
+Please ensure your internet network such as your VPN does not interfere with your build.
+
+To run container:
+
+`./run_local.sh`
```
davidread, line 32: These 3 lines are the key instruction of how to run Fala in Docker. I should move it up, out of the 'tips', so it's the main instruction of the 'Installation for development' section

README.md
```diff
@@ -10,8 +10,64 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+
+This shell script contains commands that build and runs the fala app.
+
+If you're not using `Docker Desktop`, you may want to have two terminals open for the following:
+
+1. To `exec` into the fala docker container.
```
davidread, line 38: The user will already have a terminal running `run_local.sh`, so should probably acknowledge that is the first terminal, and the `exec` one is the second.

README.md
```diff
@@ -10,8 +10,64 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+
+    `docker exec -it fala bash`
+
+    From the shell inside the container you can run some tests e.g.
+
+    `python manage.py test`
+
+2. Logs and debugging.
```
davidread, line 46: Maybe call this one just `debugging`, since the `run_local.sh` terminal already shows the logs. You still need to do this `docker attach` for pdb debugging

README.md
davidread: I think this is such a minor preference, I'm going to withdraw it

README.md
psweeting1: Moved up out of tips into `Docker` header sections.

README.md
psweeting1: Yep I agree, refactored and moved around now.

README.md
```diff
@@ -10,8 +10,58 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+
+`./run_local.sh`
+
+This shell script contains commands that build and runs the fala app.
+
+If you're not using `Docker Desktop`, you may want to have two terminals open.
+
+### Tips for developing with docker containers
+
+Code editing - You can edit the code on your local disk, with a local editor, as normal. (You don't have to edit the
```
MazOneTwoOne, line 32: I think this sentence should be this without the parenthesis:

> Code editing - You can edit the code on your local disk, with a local editor, as normal. You don't have to edit the files inside the Docker container, because your local directory is mounted into container. When you save a file, it becomes present in the container immediately, and the server restarts.

README.md
```diff
@@ -10,8 +10,58 @@ The website uses the LAA Legal Adviser API. The code is hosted at https://github
...
+To `exec` into the Fala docker container, start a new terminal and perform the following:
+
+`docker exec -it fala bash`
+
+From the shell inside the container you can run some tests e.g.
+
+`python manage.py test`
+
+Alternatively, some editors have functionality to hook into running containers, such as VS Code's 'Dev Containers' 
+extension. Docker Desktop offers a UI version of looking through logs, exec and many other interactions.
```
MazOneTwoOne, line 49: The 'Docker' extension on VS Code is much better, so might be worth updating to this

> Alternatively, some editors have functionality to hook into running containers, such as VS Code's 'Docker' 
> extension.

MazOneTwoOne (commented):
Just some suggested changes, not a blocker.

README.md
davidread: The 'Docker' extension lets you attach to a shell, saving you typing `docker attach fala`. The 'Dev Containers' extension goes further, allowing you to use VS Code's debugger. Seems reasonable to mention both.

davidread (approved):
There are little improvements suggested, but I suggest we merge this, to bank the main thrust of this PR, which is foundational to all the team's devs getting setup. And other little improvements can be in a separate PR.

PR #206, opened 2024-03-20
Pinning only in *generated* requirements files

davidread:

A new policy of unpinning the .in files - see requirements.md for full explanation of proposed change and reasoning. It makes it easier to keep dependencies up to date.

No versions are actually changed in requirements.txt, since we're on the latest in most cases (apart from django*, which are left pinned for now)

## Checklist


docs/requirements.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Updating dependencies
+
+### Automated updates - Dependabot or Snyk
+
+These automated tools generally increment the Generated files to the latest. The PRs they create should be tested, and if they work they can be merged. These automated tools don't look at the .in source files, so the constraints we put in them aren't taken account of. But if the tests fail, then the human can look in the source file's constraint to see if that already records a reason for not being able to upgrade.
+
+### Manual updates
+
+We should occasionally run `pip-compile`, to update the generated files with the latest available versions. Unlike the Dependabot/Snyk automation, it does take account of the constraints contained in the source files. So it avoids major breaking changes and known incompatibilities. And it's good for updates across all the dependencies at once. And during testing you find out if new versions are problematic, which require constraints adding to the source files.
```
exonian, line 27: > If pip-compile finds an existing requirements.txt file that fulfils the dependencies then no changes will be made, even if updates are available.

> To force pip-compile to update all packages in an existing requirements.txt, run pip-compile --upgrade.

I wonder if this flag is worth including here

exonian (commented):
I'm strongly in favour of unpinning the .in files, for the reasons you describe.

I don't super love the interaction between dependabot/snyk and our requirements setup, though. It feels like the two systems are fighting against one another, to an unknown extent of practical inconvenience that may be more just inelegance than anything, so I'll work through a couple of thoughts below:

I figure it could -- rarely -- lead to a situation where an automated upgrade takes us past a constraint and it works so we accept it (in .txt) but fail to notice that it's done so and therefore don't update the constraint (in .in), and then a future `pip-compile` respecting that constraint would effectively decrement the version again. I suppose we would just need to keep our eye out for that when we ran and committed the `pip-compile`, instead choosing to change the constraint instead of the generated version. Ok I'm quite satisfied with that, now, but I'll leave my train of thought here for the record :D

I wonder if we could or should schedule the running of `pip-compile`, perhaps as a github action creating its own PRs?

Can we come up with a strategy to keep challenging any constraints that we do have in our .in files? Could/should this be a scheduled task? Or do the two points above achieve this? ie letting a bot take us past a constraint by updating our generated requirements directly, and then routinely regenerating with `pip-compile` and revising the constraints if they've already been surpassed. The alternative, and what came to mind first, was some sort of process where we remove the constraints, run `pip-compile`, and see if it works, but maybe we don't need this if depedabot/snyk are driving us forwards.

starswan (commented):
I'm in favour of this too - it mirrors what we do in Ruby with a Gemfile, where the major release is pinned (or often not) but the minor release is only pinned if there is a known incompatibility.

docs/requirements.md
davidread: Much appreciated. I've included this flag now.

davidread:
@exonian 
> would effectively decrement the version again
Good thinking - I've added a note to watch out for this

> I wonder if we could or should schedule the running of pip-compile
I'm warm to this, but it might not be worth it just for this little repo. It feels a bit like Dependabot/Snyk should do this, if it was important and easy. One to ponder

> Can we come up with a strategy to keep challenging any constraints
Same. Good thinking, but one to ponder

requirements/generated/requirements-dev.txt
```diff
@@ -65,16 +65,14 @@ nodeenv==1.8.0
     # via pre-commit
 packaging==24.0
     # via build
-pip-tools==7.4.1
+pip-tools==6.13.0
```
psweeting1, line 68: Downgrading pip tools, was there a problem with the reason being to high?

psweeting1: Never mind, I found the reasoning.

psweeting1 (approved):
LGTM

PR #217, opened 2024-03-22
Python version 3.12 - add docs to install this particular version

davidread:

We need folks to use Python 3.12, not any other version, since otherwise CircleCI will raise errors on the pip-compile step, because the python version is in the generated requirements*.txt files.

pyenv is the best choice, since other CLA repos have a range of versions.

When a user has installed pyenv, then when they run `python`, it runs the pyenv shim instead. If they are currently in the fala directory, then the shim sees the `.python-version` file and [uses the python version specified in that](https://github.com/pyenv/pyenv?tab=readme-ov-file#understanding-python-version-selection).

Intentionally left blank.

## Checklist


docs/virtual-env.md
```diff
@@ -2,8 +2,39 @@
...
+3. To make the shell changes take effect:
+
+        exec "$SHELL"
+
+    (or alternatively, restart your shell)
+
+4. Install into pyenv the python version this repo uses (which is defined in `.python-version`):
+
+        pyenv install 3.12 --skip-existing
+
```
MazOneTwoOne, line 24: I would add this step. 

```
5. Select the 'Pyenv-installed' Python as the version to use, by running the following command:

`pyenv local 3.12`
```

davidread: Good call about adding a version check. `pyenv local` will change `.python-version` though, which is probably not what's needed here. The stumbling block I've encountered is if you've created a venv in one python version, it's hard to switch the version, so I've added a paragraph on this. But if there is another pain point here, let me know

MazOneTwoOne (approved):

PR #221, opened 2024-03-27
EL-1451 feature flag

psweeting1:

Adding a feature flag for our up and coming FALA development upgrade work. 

Though "false" is used as a boolean in the kubenetes deployment.yml, it can be anything we like, it can even be an empty. All we care about if the string true is visible or not.

## Checklist

- [EL-1451](https://dsdmoj.atlassian.net/browse/EL-1451)

[EL-1451]: https://dsdmoj.atlassian.net/browse/EL-1451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ docs/_build
 env
 env*
 venv
+.venv
 
 # intellij
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,6 @@ docs/_build
 env
 env*
 venv
-.venv
 
 # intellij
 .idea/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,5 +15,6 @@ services:
     environment:
       LAALAA_API_HOST: "https://laa-legal-adviser-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk"
       DJANGO_SETTINGS_MODULE: "fala.settings.local"
+      FALA_NO_MAP: "fala.settings.local"
     ports:
       - "8013:8000"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,5 @@ services:
     environment:
       LAALAA_API_HOST: "https://laa-legal-adviser-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk"
       DJANGO_SETTINGS_MODULE: "fala.settings.local"
-      FALA_NO_MAP: "fala.settings.local"
     ports:
       - "8013:8000"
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -38,6 +38,7 @@ docs/_build
 env
 env*
 venv
+.venv
```
davidread, line 41: What's wrong with `venv` ;)

fala/settings/base.py
```diff
@@ -214,6 +214,8 @@ def root(*x):
 
 LAALAA_API_HOST = os.environ.get("LAALAA_API_HOST", None)
 
+FALA_NO_MAP = os.environ.get("FALA_N0_MAP", "").lower() == "true"
```
davidread, line 217: I think you have a zero instead of an O `FALA_N0_MAP`

davidread: I think it's worth adding prefix `FEATURE_FLAG_` to feature flags, e.g. `FEATURE_FLAG_NO_MAP`, which communicates that it should be removed once the feature is released

davidread: Nice plan to default it to False

docker-compose.yml
```diff
@@ -15,5 +15,6 @@ services:
     environment:
       LAALAA_API_HOST: "https://laa-legal-adviser-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk"
       DJANGO_SETTINGS_MODULE: "fala.settings.local"
+      FALA_NO_MAP: "fala.settings.local"
```
davidread, line 18: This doesn't look right to me - I believe it will set the environment variable to `fala.settings.local`, which is not what we're after. (Although things will still work because Django will take in preference the value from local.py, so it will still work if the user created that file). So I suggest we omit this line, and just let the user edit the value in local.py.

.gitignore
psweeting1: Ah good point, in the readme for the virtual env install it's `venv`. I would argue to make a change in a quick PR to `.venv` as this is what we use in our Python Flask work and exploratory work. 

It's very a minor thing, but we should pick one or another :)

davidread: I'm easy. But yes, it's right to take it out of this PR

fala/settings/base.py
psweeting1: I honestly don't know how a zero got in there!

docker-compose.yml
psweeting1: Yeah, after doing some testing on a different branch, I can confirm it's not needed.

davidread (approved):

PR #229, opened 2024-04-05
EL-1431: Search page - Update styling

MazOneTwoOne:

- Updating FALA landing / search page with new design changes

- Not able to create a 3 grid column for checkboxes, using existing GDS layouts and styling;
  - We could do this but would end up having to make our own CCS classes, which is not ideal.
- The checkboxes backgrounds are not white and instead transparent, as per their design;
  - There is an open issue discussing this [Why do checkboxes and radios have transparent backgrounds? · Issue #1625 · alphagov/govuk-frontend](https://github.com/alphagov/govuk-frontend/issues/1625) 
  - There is a [workaround ](https://github.com/daniel-ac-martin/NotGovUK/pull/657/files) which I have borrowed. 

## Checklist


fala/apps/adviser/views.py
```diff
@@ -3,11 +3,18 @@
...
+    from .forms_old import AdviserSearchForm
 
 
 class AdviserView(TemplateView):
-    template_name = "adviser/search.html"
+    # https://docs.djangoproject.com/en/5.0/topics/class-based-views/#usage-in-your-urlconf - this is how `template_name`
+    if settings.FEATURE_FLAG_NO_MAP:
+        template_name = "adviser/search.html"
+    else:
+        template_name = "adviser/search-old.html"
```
psweeting1, line 17: `search_old` I get template not found when feature flag is false.

fala/apps/adviser/forms.py
```diff
@@ -27,14 +27,19 @@ class AdviserSearchForm(forms.Form):
...
         max_length=30,
-        help_text=_("For example, <span class='notranslate' translate='no'>SW1H 9AJ</span>"),
+        help_text=_("For example, <span class='notranslate' translate='no'>SW1H</span>"),
         required=False,
         widget=FalaTextInput(),
     )
 
-    name = forms.CharField(label=_("Organisation name"), max_length=100, required=False, widget=FalaTextInput())
+    name = forms.CharField(
+        label=_("<label class='govuk-label govuk-label--s'>Organisation name</label>"),
```
psweeting1, line 38: Again Label HTML in Label HTML. Just pass the string `Organisation`

fala/apps/adviser/forms.py
```diff
@@ -27,14 +27,19 @@ class AdviserSearchForm(forms.Form):
     page = forms.IntegerField(required=False, widget=forms.HiddenInput())
 
     postcode = forms.CharField(
-        label=_("Enter postcode"),
+        label=_("<label class='govuk-label govuk-label--s'>Postcode</label>"),
         max_length=30,
-        help_text=_("For example, <span class='notranslate' translate='no'>SW1H 9AJ</span>"),
+        help_text=_("For example, <span class='notranslate' translate='no'>SW1H</span>"),
```
psweeting1, line 32: I know this a copy. They're being a little naughty here. While they're using inline HTML here, which isn't great, the way they set up the marcos in form.html makes working with this a bit of a pain, so another solution here is to use `django.utils.safestring`

```
from django.utils.safestring import mark_safe
help_text = mark_safe("<label class='govuk-label govuk-label--s'>Postcode</label>")
```

fala/apps/adviser/forms.py
```diff
@@ -27,14 +27,19 @@ class AdviserSearchForm(forms.Form):
     page = forms.IntegerField(required=False, widget=forms.HiddenInput())
 
     postcode = forms.CharField(
-        label=_("Enter postcode"),
+        label=_("<label class='govuk-label govuk-label--s'>Postcode</label>"),
```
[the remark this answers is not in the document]
psweeting1: Within macros `form.html` it already sets the field label into a HTML label.

psweeting1:
All FALA end to end tests are failing because there are HTML changes. Something we didn't think would be an issue. 


If we look at the step here. We can see it's using xpath to find the header, however the pages structure has changed so this xpath is now wrong. 
```
@step("I am on the Find a legal aid adviser homepage")
def step_impl_homepage(context):
    homepage_url = f"{CLA_FALA_URL}"
    context.helperfunc.open(homepage_url)
    title_xpath = context.helperfunc.find_by_xpath(
        "//html/body/div/main/div/div/h1"
    ).text.replace("\n", " ")

    assert title_xpath == FALA_HEADER
  ```
  
  We can use a common_step method called `assert_header_on_page` instead and assert it equals to FALA_HEADER.

File: fala/templates/adviser/search.html [hunk 5]
```diff
@@ -7,70 +7,88 @@
...
+  <div class="govuk-grid-column-two-thirds">
+      <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
+      <h1 class="govuk-heading-xl"> Find a legal aid adviser or family mediator </h1>
+      <p class="govuk-body">You need to provide information about your finances to find out if you qualify for legal aid.</p>
+      <p class="govuk-body"><a class="govuk-link" href="https://www.gov.uk/check-legal-aid" target="_blank" rel="noopener">Check if you qualify for legal aid (opens in a new tab)</a></p>
+  </div>
 
-      {% set errorMessage = {'err': ""} %}
+  {% block search_form %}
+      <div class="govuk-grid-column-full" style="background-color: #f3f2f1; padding: 1em; margin: 0.5em;">
```
psweeting1, line 19: Inline CSS needs to be removed. This should be a CSS class that can be re-used throughout the project.

fala/templates/adviser/search.html
```diff
@@ -7,70 +7,88 @@
...
-          govuk-form-group--error
-        {% endif %}
-        "
-      id="fala_questions">
+{% block content %}
+  <div class="govuk-grid-column-two-thirds">
+      <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
+      <h1 class="govuk-heading-xl"> Find a legal aid adviser or family mediator </h1>
+      <p class="govuk-body">You need to provide information about your finances to find out if you qualify for legal aid.</p>
+      <p class="govuk-body"><a class="govuk-link" href="https://www.gov.uk/check-legal-aid" target="_blank" rel="noopener">Check if you qualify for legal aid (opens in a new tab)</a></p>
```
psweeting1, line 15: URL to `check-legal-aid` should be in `base.py` and set up as a variable which can be accessed by the template. 

This allows the URL to be used in other places in the template and if any changes are made, you only need to change it one location. 

base.py
`CHECK_LEGAL_AID_URL = "https://www.gov.uk/check-legal-aid"`

Jinja Template:
`<p class="govuk-body"><a class="govuk-link" href="{CHECK_LEGAL_AID_URL}" target="_blank" rel="noopener"></p>`

fala/apps/adviser/forms.py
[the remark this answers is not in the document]
MazOneTwoOne: @psweeting1 - choose to do this way as we are using `forms.py` (with feature flag) and `forms_old.py` (without feature flag. 

Can I use have a `forms.html` (with feature flag) and `forms_old.html` (without feature flag), for the macros??

fala/apps/adviser/forms.py
MazOneTwoOne: @psweeting1 - confused by this comment - reason not passing a string is because I want the styling, to be different. Would I achieve this by adding the styling to the `forms.html` macro??

fala/templates/adviser/search.html
MazOneTwoOne: @psweeting1 that doesn't work.

fala/templates/adviser/search.html
MazOneTwoOne: @psweeting1 so my understanding of `gov.uk` classes, is that you can't override a base class. So if I made another class, it won't apply to the `gov.uk` base class. I can only amend using `gov.uk override classes` or as I've done above, add in additional stylings.

fala/templates/adviser/search.html
MazOneTwoOne: should have been this - href="{{ CHECK_LEGAL_AID_URL }}"

fala/apps/adviser/views.py
```diff
@@ -22,6 +26,8 @@ def get(self, request, *args, **kwargs):
                 "current_url": current_url,
                 "GOOGLE_MAPS_API_KEY": settings.GOOGLE_MAPS_API_KEY,
                 "LAALAA_API_HOST": settings.LAALAA_API_HOST,
+                "FEATURE_FLAG_NO_MAP": settings.FEATURE_FLAG_NO_MAP,
```
starswan, line 29: As previously noted, we probably shouldn't expose this to the template (and I think its unused)

MazOneTwoOne: have now removed.

fala/urls.py
```diff
@@ -2,7 +2,7 @@
 from django.conf import settings
 from django.conf.urls.static import static
 from django.urls import re_path
-from adviser.views import AdviserView, AccessibilityView
+from fala.apps.adviser.views import AdviserView, AccessibilityView
```
exonian, line 5: I am surprised this is necessary. `adviser` is included in the [project's `INSTALLED_APPS`](https://github.com/ministryofjustice/fala/blob/master/fala/settings/base.py#L188-L190) and therefore should be available to import from by name. May be worth reverting this line and seeing if anything actually breaks, if it was added while trying other stuff out and if you haven't already.

MazOneTwoOne: @exonian you are indeed correct that by removing this, that nothing will break. 

I made the change as my IDE (VS Code) was complaining to me, that there were missing imports. 
Screenshot 2024-04-17 at 16 59 55

Having the long path, stop the complaints.

exonian: Yeah I realised when I saw the same on Stephen's PR that it was likely IDEs. I wonder if there's a Django plugin that can help it understand that installed apps are on the Python path 🤔

MazOneTwoOne: Potentially - I'll remove/revert this change in the mean time.

fala/apps/adviser/forms.py
```diff
@@ -26,16 +28,30 @@ def __init__(self, attrs={}):
 class AdviserSearchForm(forms.Form):
     page = forms.IntegerField(required=False, widget=forms.HiddenInput())
 
-    postcode = forms.CharField(
-        label=_("Enter postcode"),
-        max_length=30,
-        help_text=_("For example, <span class='notranslate' translate='no'>SW1H 9AJ</span>"),
+    if settings.FEATURE_FLAG_NO_MAP:
```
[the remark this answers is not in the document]
MazOneTwoOne: @exonian so this is only an if/else feature flag statement on the `postcode` form filed (change to name and helptext). So I think this already covers your request to keep it maintainable?? 

Might look like more has changed, as I did some formatting on how the form fields were laid out in this class, to keep them consistent.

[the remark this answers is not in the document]
exonian: Yeah all I mean is that instead of both parts of the conditional writing out the whole field definition, which is several lines of which most are the same for both, it's possible to write the field once and then the conditional only changes the two lines it needs to change.
For me, that feels more maintainable (if we were to keep this code long term), because it's a lot easier to see what the `if` statement is doing if things like max_length and widget aren't in both the `if` and the `else` clauses.
It would look something like:
`
postcode = forms.CharField(
    label=_("Enter postcode"),
    max_length=30,
    help_text=_("For example, <span class='notranslate' translate='no'>SW1H 9AJ</span>"),
    required=False,
    widget=FalaTextInput(),
)

def __init__(self, *args, **kwargs):
    super()
    if not settings.FEATURE_FLAG_NO_MAP:
        self.fields['postcode'].label = _("Postcode")
        self.fields['postcode'].help_text=_(mark_safe("For example, <span class='notranslate' translate='no'>SW1H</span>"))
`

I don't mind us not, here, just wanted to point out this way is available and will sometimes be what we need. To my eye, it's neater, but it lacks simplicity so I get it.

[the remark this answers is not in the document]
MazOneTwoOne: @exonian This makes sense. Have now changed.

exonian (approved):
I can't deny my eyes glazed over slightly for the html, but nothing jumped out, and the Python looks good

PR #232, opened 2024-04-11
EL-1434 new results page with pagination

starswan:

New results page with pagination

## Checklist

### DIFF
--- a/fala/apps/adviser/forms.py
+++ b/fala/apps/adviser/forms.py
@@ -91,6 +91,10 @@ class AdviserSearchForm(forms.Form):
                 self.add_error("postcode", "%s %s" % (_(" ".join((msg1, region))), _(msg2)))
         return data
 
+    @property
+    def current_page(self):
+        return self.cleaned_data.get("page", 1)
+
     def search(self):
         if self.is_valid():
             try:
--- a/fala/apps/adviser/laa_laa_paginator.py
+++ b/fala/apps/adviser/laa_laa_paginator.py
@@ -1,0 +1,47 @@
+class LaaLaaPaginator(object):
+    def __init__(self, count, page_size, window_size, current_page):
+        self._window_size = window_size
+        if current_page is None:
+            self._current_page = 1
+        else:
+            self._current_page = current_page
+        self._very_last_page = count // page_size + 1
+
+    @property
+    def page_range(self):
+        return range(self._first_page_num(), self._last_page_num() + 1)
+
+    def _first_page_num(self):
+        if self._current_page > self._window_size:
+            return self._current_page - self._window_size
+        else:
+            return 1
+
+    def _last_page_num(self):
+        if self._current_page + self._window_size < self._very_last_page:
+            return self._current_page + self._window_size
+        else:
+            return self._very_last_page
+
+    def current_page(self):
+        class LaaLaaPage(object):
+            def __init__(self, page_num, last_page):
+                if page_num is None:
+                    self._page_num = 1
+                else:
+                    self._page_num = page_num
+                self._last_page = last_page
+
+            def has_previous(self):
+                return self._page_num > 1
+
+            def has_next(self):
+                return self._page_num < self._last_page
+
+            def previous_page_number(self):
+                return self._page_num - 1
+
+            def next_page_number(self):
+                return self._page_num + 1
+
+        return LaaLaaPage(self._current_page, self._last_page_num())
--- a/fala/apps/adviser/tests/test_laa_laa_paginator.py
+++ b/fala/apps/adviser/tests/test_laa_laa_paginator.py
@@ -1,0 +1,45 @@
+from django.test import SimpleTestCase
+
+from fala.apps.adviser.laa_laa_paginator import LaaLaaPaginator
+
+
+class LaaLaaPaginatorTest(SimpleTestCase):
+    def test_short_data_has_one_page(self):
+        pag = LaaLaaPaginator(count=5, page_size=10, window_size=3, current_page=1)
+        self.assertEqual(1, len(pag.page_range))
+
+    def test_short_data_has_no_previous(self):
+        pag = LaaLaaPaginator(count=5, page_size=10, window_size=3, current_page=1).current_page()
+        self.assertFalse(pag.has_previous())
+
+    def test_short_data_has_no_next(self):
+        pag = LaaLaaPaginator(count=5, page_size=10, window_size=3, current_page=1).current_page()
+        self.assertFalse(pag.has_next())
+
+    def test_medium_data_has_two_pages(self):
+        pag = LaaLaaPaginator(count=15, page_size=10, window_size=3, current_page=1)
+        self.assertEqual(2, len(pag.page_range))
+
+    def test_medium_data_page_1_paginates(self):
+        page_1 = LaaLaaPaginator(count=15, page_size=10, window_size=3, current_page=1).current_page()
+        self.assertEqual([False, True], [page_1.has_previous(), page_1.has_next()])
+
+    def test_medium_data_page_2_paginates(self):
+        page_2 = LaaLaaPaginator(count=15, page_size=10, window_size=3, current_page=2).current_page()
+        self.assertEqual([True, False], [page_2.has_previous(), page_2.has_next()])
+
+    def test_large_data_has_correct_pages(self):
+        pag = LaaLaaPaginator(count=105, page_size=10, window_size=2, current_page=1)
+        self.assertEqual(range(1, 4), pag.page_range)
+
+    def test_large_data_page_2_pages(self):
+        pag = LaaLaaPaginator(count=105, page_size=10, window_size=2, current_page=2)
+        self.assertEqual(range(1, 5), pag.page_range)
+
+    def test_large_data_page_3_pages(self):
+        pag = LaaLaaPaginator(count=105, page_size=10, window_size=2, current_page=3)
+        self.assertEqual(range(1, 6), pag.page_range)
+
+    def test_large_data_page_5_pages(self):
+        pag = LaaLaaPaginator(count=105, page_size=10, window_size=2, current_page=5)
+        self.assertEqual(range(3, 8), pag.page_range)
--- a/fala/apps/adviser/tests/test_search_view_function.py
+++ b/fala/apps/adviser/tests/test_search_view_function.py
@@ -1,3 +1,4 @@
+import bs4
 from django.test import SimpleTestCase, Client, override_settings
 from django.urls import reverse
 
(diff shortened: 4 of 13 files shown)


fala/apps/adviser/tests/test_laa_laa_paginator.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+
+from fala.apps.adviser.laa_laa_paginator import LaaLaaPaginator
+
+class LaaLaaPaginatorTest(unittest.TestCase):
```
exonian, line 5: I would advocate for always using a Django testcase subclass. For stuff using a database, it will be more critical, and here we don't need stuff like the transactional behaviour of `django.test.TestCase`, but I think using [`django.test.SimpleTestCase`](https://docs.djangoproject.com/en/5.0/topics/testing/tools/#simpletestcase) is best here

I believe the django test runner would still find and run `unittest.TestCase` tests just fine, but using `django.test.SimpleTestCase` gives us some useful utilities like setup and teardown hooks (although for these tests we don't need those, admittedly) and -- most importanly -- consistency with other applications where we have a db and will definitely want to use Django subclasses of TestCase.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,130 @@
...
+              {% if item.organisation.website %}
+                <p class="govuk-body">
+                  <span>{{ _('Website') }}:</span>
+                  <a class="govuk-link" href="{{ item.organisation.website|human_to_url }}" class="url" target="_blank" rel="noopener">
+                    {{ item.organisation.website|url_to_human + " (Opens in new tab)" }}
+                  </a>
+                </p>
+              {% endif %}
+              {% if item.categories|length %}
+                <h4 class="govuk-heading-s govuk-!-margin-bottom-2">{{ _('Categories of law') }}</h4>
```
psweeting1, line 64: You're using a H4 here. There is no H3 anywhere in the DOM, so the header level order doesn't make sense. I don't think this should be a header anyways as the header level within this loop will be all over the place. This should be a `span` to fix that.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,135 @@
...
+        {% endif %}
+      {% else %}
+        {% trans result_count=data.count, page_count=data.results|length, org_name=form.name.value() %}
+          Showing {{ page_count }} result for <strong>{{ org_name }}</strong>.
+          {% pluralize %}
+          {{ result_count }} results in total. Showing {{ page_count }} results for <strong>{{ org_name }}</strong>.
+        {% endtrans %}
+      {% endif %}
+      <hr/>
+      <div class="org-list">
```
psweeting1, line 28: This class isn't needed. In the CSS the rules are the following for this.

`.find-legal-adviser .search-results-list .org-list `

This selector targets elements with the class "find-legal-adviser" that have descendants elements of the following order above.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,135 @@
...
+        {% for item in data.results %}
+          <div class="org-list-item results-list-item" data-lat="{{ item.location.point.coordinates.1 }}" data-lon="{{ item.location.point.coordinates.0 }}" data-id="{{ loop.index }}">
```
psweeting1, line 30: Same here with `org-list-item` this CSS class is not used and never triggered.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,135 @@
...
+            <div class="org-details">
```
psweeting1, line 31: Sensing a theme here. `org-details` is not used.

fala/templates/adviser/results.html
psweeting1: Actually you don't need this `div` at all now, so you can delete this HTML.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,135 @@
...
+              {% if item.distance %}
+                <p class="miles-away govuk-body">
+                  <span class="govuk-visually-hidden">{{ _('Distance') }}</span>
+                  {% trans miles=item.distance|round(2) %}{{ miles }} miles away{% endtrans %}
+                </p>
+              {% endif %}
+              <span class="fn org govuk-body-l govuk-!-font-weight-bold">{{ item.organisation.name }}</span>
```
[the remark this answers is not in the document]
psweeting1: I had a good chat with Maz around headers in HTML, but we agreed this needs to be a H2 and not a span as it is a header.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,135 @@
...
+              {% if item.organisation.website %}
+                <p class="govuk-body">
+                  <span>{{ _('Website') }}:</span>
+                  <a class="govuk-link" href="{{ item.organisation.website|human_to_url }}" class="url" target="_blank" rel="noopener">
+                    {{ item.organisation.website|url_to_human + " (Opens in new tab)" }}
+                  </a>
+                </p>
+              {% endif %}
+              {% if item.categories|length %}
+                <span class="govuk-heading-s govuk-!-margin-bottom-2">{{ _('Categories of law') }}</span>
```
psweeting1, line 61: I really sorry, for `Categories of law` header, I told you to make this a span HTML. But actually it can be a H3

fala/apps/adviser/forms.py
```diff
@@ -92,6 +96,7 @@ def search(self):
                 if "error" in data:
                     self.add_error("postcode", (data["error"]))
                     return {}
+                print("returning data", data)
```
psweeting1, line 99: Hello print, what brings you here.

fala/assets-src/sass/main.scss
```diff
@@ -187,3 +187,11 @@ body.fala-dev {
   margin-top:1em;
   clear:both;
 }
+
+.miles-away {
+  color: $govuk-secondary-text-colour;
+}
+
+.results-list-item {
```
psweeting1, line 195: **Optional**, I'm not going to lose sleep over this because it's easy to refactor. 

I would follow BEM here for future proofing: https://getbem.com/

So you can do:

```
.results-list
&__item {
margin-top: govuk-spacing(4);
}
```

This allows extensions and future styling for the results list items or even something else on the result page.

fala/templates/adviser/search.html
```diff
@@ -8,7 +8,7 @@
 {% import "macros/forms.html" as Form %}
 
 {% block search_form %}
-  <form action="" class="laa-fala__search" novalidate>
+  <form action="/search" class="laa-fala__search" novalidate>
```
exonian, line 11: Doesn't this need to be feature flagged?

fala/apps/adviser/views.py
```diff
@@ -30,3 +35,36 @@ def get(self, request, *args, **kwargs):
 
 class AccessibilityView(TemplateView):
     template_name = "adviser/accessibility-statement.html"
+
+
+def fala_search(request):
```
exonian, line 40: What's the rationale for mixing in a function-based view alongside our class-based views?

fala/apps/adviser/laa_laa_paginator.py
```diff
@@ -0,0 +1,47 @@
+class LaaLaaPaginator(object):
```
exonian, line 1: What does this do that we needed to write our own instead of using Django's paginator like we discussed?

starswan: Django's paginator doesn't work here. The items are already paginated - Django assumes you have an array to be paginated

fala/apps/adviser/views.py
starswan: Now a class based version

fala/apps/adviser/views.py
```diff
@@ -37,3 +38,36 @@ def get(self, request, *args, **kwargs):
...
+            pages = LaaLaaPaginator(self._data["count"], 10, 3, self._form.current_page)
+            params = {
+                "postcode": self._form.cleaned_data["postcode"],
+                "name": self._form.cleaned_data["name"],
+            }
+            categories = list(map(lambda c: "categories=" + c, self._form.cleaned_data["categories"]))
+            context.update(
+                {"form": self._form, "data": self._data, "pages": pages, "params": params, "categories": categories}
+            )
+        else:
```
exonian, line 71: I have a mild preference for doing the common bits outside of the conditional, and having the conditional only do what's different.
In this case that would look something like
```
context.update(form=self._form, data=self._data)  # this isn't different from passing them as a dict

if self._data:
    <snip>
    context.update(pages=pages, params=params, categories=categories)
```

fala/apps/adviser/tests/test_adviser_extras.py
```diff
@@ -1,6 +1,6 @@
 import unittest
 
-from adviser.templatetags import adviser_extras
+from fala.apps.adviser.templatetags import adviser_extras
```
exonian, line 3: I think my comment got lost: this shouldn't be necessary and is an artefact of the IDE not having a proper Django runtime

fala/apps/adviser/tests/test_search_view_function.py
```diff
@@ -17,6 +17,42 @@ def test_invalid_postcode_generates_error(self):
...
+
+    def test_category_search_has_good_result(self):
+        response = self.client.get(self.url, self.data)
+        self.assertEqual(200, response.status_code)
+
+    def test_category_search_has_just_user_input(self):
+        response = self.client.get(self.url, self.data)
+        self.assertNotContains(response, "PE30 1AA")
+
+    def test_category_search_heading_closeness(self):
```
exonian, line 35: It is not clear to me from these tests what they're intending to achieve. Are they just tests that particular strings are in the template, or is there logic that controls those strings? If the latter, my instinct is we should have counter-tests to demonstrate when they should be absent, otherwise I'm unsure of their value

fala/apps/adviser/views.py
```diff
@@ -37,3 +38,36 @@ def get(self, request, *args, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self._data:
+            pages = LaaLaaPaginator(self._data["count"], 10, 3, self._form.current_page)
+            params = {
+                "postcode": self._form.cleaned_data["postcode"],
+                "name": self._form.cleaned_data["name"],
+            }
+            categories = list(map(lambda c: "categories=" + c, self._form.cleaned_data["categories"]))
```
exonian, line 67: I would be lying if I said I could read this line 😅

fala/templates/adviser/search.html
```diff
@@ -18,7 +18,7 @@ <h1 class="govuk-heading-xl"> Find a legal aid adviser or family mediator </h1>
   {% block search_form %}
       <div class="govuk-grid-column-full">
         <div class="laa-fala__grey-box">
-          <form action="" novalidate>
+          <form action="search" novalidate>
```
exonian, line 21: Has this leaked out of the feature flag again? Or is this existing template name now used only by the flagged version, and the template that was here before is now called old_search?

fala/urls.py
```diff
@@ -2,12 +2,13 @@
 from django.conf import settings
 from django.conf.urls.static import static
 from django.urls import re_path
-from adviser.views import AdviserView, AccessibilityView
+from fala.apps.adviser.views import AccessibilityView, AdviserView, SearchView
```
exonian, line 5: As above

fala/apps/adviser/views.py
```diff
@@ -37,3 +38,36 @@ def get(self, request, *args, **kwargs):
 
 class AccessibilityView(TemplateView):
     template_name = "adviser/accessibility-statement.html"
+
+
+class SearchView(ListView):
```
exonian, line 43: I can make my peace with this view being unorthodox in that it's very much not single-function and therefore has weirdness like different templates for which thing it's doing. As we discussed, the inability to separate validation from data-fetching means it's not possible to have a normal split between form-handling view (that validates) and results-showing view (that may call the API to do so).

However, this code needs some comments, possibly just in `get_context_data()` but probably in at least `get_template_names()` too, to allow devs to understand the logical flow through this view, because it is non-obvious.

fala/templates/adviser/results.html
```diff
@@ -0,0 +1,129 @@
...
+        {% for item in data.results %}
+          <li class="results-list-item">
+            {% if item.distance %}
+              <p class="miles-away govuk-body">
+                <span class="govuk-visually-hidden">{{ _('Distance') }}</span>
+                {% trans miles=item.distance|round(2) %}{{ miles }} miles away{% endtrans %}
+              </p>
+            {% endif %}
+            <h2 class="govuk-body-l govuk-!-font-weight-bold">{{ item.organisation.name }}</h2>
+            <div class="telephone">
```
psweeting1, line 35: Sorry I found this CSS class isn't being used for anything (Not even in the stylesheets). So this can go bye bye!

fala/templates/adviser/search.html
starswan: This is now the new search, so the flag is already on for this one

fala/apps/adviser/views.py
starswan: I've added a comment to help

fala/apps/adviser/tests/test_search_view_function.py
starswan: added some extra contra-tests

fala/apps/adviser/views.py
starswan: I've added some extra comments. In EL-1433 (where there is more logic around the postcodes) we managed to improve this

fala/apps/adviser/views.py
exonian: It doesn't help a lot, I'm afraid. I still struggle to predict what this output even looks like 🥴 
Is it like `["categories=a", "categories=b"]`?

If I can understand what it's doing, I may be able to suggest a more readable way of doing it

fala/apps/adviser/tests/test_search_view_function.py
exonian: They've not really done a lot for readability, have they?

fala/apps/adviser/views.py
exonian: OK great this is good for now and we can pick it up further there

fala/apps/adviser/tests/test_search_view_function.py
starswan: managed to come up with a slightly better solution

fala/assets-src/sass/main.scss
```diff
@@ -215,7 +215,29 @@ body.fala-dev {
...
   }
 }
+
+.miles-away {
+  color: $govuk-secondary-text-colour;
+}
+
+.results-list-item {
+  margin-top: govuk-spacing(4);
+  border-top: 1px solid #b1b4b6;
```
psweeting1, line 229: Is `#b1b4b6` not the same colour as `govuk-colour("light-grey")`

MazOneTwoOne: It would be the same as `govuk-colour("mid-grey")` (see image below, [here is the link](https://design-system.service.gov.uk/styles/colour/))


Screenshot 2024-04-23 at 09 02 45

psweeting1 (approved):
LGTM

PR #240, opened 2024-04-17
El 1433 crown dependencies results

starswan:

Add crown dependencies and Northern Ireland / Scotland results page

Intentionally left blank.

## Checklist

### DIFF
--- a/fala/apps/adviser/forms.py
+++ b/fala/apps/adviser/forms.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+
 from django import forms
 from django.utils.translation import gettext_lazy as _
 from django.utils.safestring import mark_safe
@@ -6,6 +7,8 @@ from django.conf import settings
 
 import laalaa.api as laalaa
 import re
+
+from .regions import Region, SCOTTISH_PREFIXES
 
 SEARCH_TYPE_CHOICES = [("location", _("Location")), ("organisation", _("Organisation"))]
 
@@ -26,6 +29,15 @@ class FalaTextInput(forms.TextInput):
 
 
 class AdviserSearchForm(forms.Form):
+    # Only used with FEATURE_FLAG_NO_MAP False
+    REGION_NAMES = {
+        Region.NI: "Northern Ireland. ",
+        Region.IOM: "the Isle of Man. ",
+        Region.JERSEY: "Jersey. ",
+        Region.GUERNSEY: "Guernsey. ",
+        Region.SCOTLAND: "Scotland. ",
+    }
+
     page = forms.IntegerField(required=False, widget=forms.HiddenInput())
 
     postcode = forms.CharField(
@@ -71,29 +83,53 @@ class AdviserSearchForm(forms.Form):
         if not data.get("postcode") and not data.get("name"):
             raise forms.ValidationError(_("Enter a postcode or an organisation name"))
         else:
-            postcodeNoSpace = data.get("postcode").replace(" ", "")
-            msg1 = "This service does not cover "
-            msg2 = "Try a postcode, town or city in England or Wales."
-            region = "England or Wales. "
-            if re.search(r"^BT[0-9]", postcodeNoSpace, flags=re.IGNORECASE):
-                region = "Northern Ireland. "
-            elif re.search(r"^IM[0-9]", postcodeNoSpace, flags=re.IGNORECASE):
-                region = "the Isle of Man. "
-            elif re.search(r"^JE[0-9]", postcodeNoSpace, flags=re.IGNORECASE):
-                region = "Jersey. "
-            elif re.search(r"^GY[1][0]", postcodeNoSpace, flags=re.IGNORECASE):
-                region = "Sark or Guernsey. "
-            elif re.search(r"^GY[9]", postcodeNoSpace, flags=re.IGNORECASE):
-                region = "Alderney or Guernsey. "
-            elif re.search(r"^GY[0-8]", postcodeNoSpace, flags=re.IGNORECASE):
-                region = "Guernsey. "
-            if region != "England or Wales. ":
-                self.add_error("postcode", "%s %s" % (_(" ".join((msg1, region))), _(msg2)))
+            if settings.FEATURE_FLAG_NO_MAP:
+                postcode = data.get("postcode")
+                if postcode and self.region == Region.ENGLAND_OR_WALES and not self._valid_postcode(postcode):
+                    self.add_error("postcode", (_("Enter a valid postcode")))
+            else:
+                region = self.region
+                if region != Region.ENGLAND_OR_WALES:
+                    region_error = self.REGION_NAMES[region]
+                    msg1 = "This service does not cover "
+                    msg2 = "Try a postcode, town or city in England or Wales."
+                    self.add_error("postcode", "%s %s" % (_(" ".join((msg1, region_error))), _(msg2)))
         return data
+
+    @property
+    def region(self):
+        postcodeNoSpace = self.cleaned_data.get("postcode").replace(" ", "")
+        if re.search(r"^BT[0-9]", postcodeNoSpace, flags=re.IGNORECASE):
+            return Region.NI
+        elif re.search(r"^IM[0-9]", postcodeNoSpace, flags=re.IGNORECASE):
+            return Region.IOM
+        elif re.search(r"^JE[0-9]", postcodeNoSpace, flags=re.IGNORECASE):
+            return Region.JERSEY
+        elif re.search(r"^GY[1][0]", postcodeNoSpace, flags=re.IGNORECASE):
+            return Region.GUERNSEY
+        elif re.search(r"^GY[9]", postcodeNoSpace, flags=re.IGNORECASE):
+            return Region.GUERNSEY
+        elif re.search(r"^GY[0-8]", postcodeNoSpace, flags=re.IGNORECASE):
+            return Region.GUERNSEY
+        elif postcodeNoSpace[:2] in SCOTTISH_PREFIXES:
+            return Region.SCOTLAND
+        else:
+            return Region.ENGLAND_OR_WALES
 
     @property
     def current_page(self):
         return self.cleaned_data.get("page", 1)
+
+    def _valid_postcode(self, postcode):
+        try:
+            data = laalaa.find(
+                postcode=postcode,
+                page=1,
+            )
+            return "error" not in data
+
+        except laalaa.LaaLaaError:
+            return False
 
     def search(self):
         if self.is_valid():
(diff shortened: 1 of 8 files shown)


fala/templates/adviser/search.html
```diff
@@ -18,6 +18,20 @@ <h1 class="govuk-heading-xl"> Find a legal aid adviser or family mediator </h1>
...
+              </h2>
+              <div class="govuk-error-summary__body">
+                <ul class="govuk-list govuk-error-summary__list">
+                  {{ Form.render_field_errors_for_summary(form.postcode) }}
+                  {{ Form.render_field_errors_for_summary(form.name) }}
+                </ul>
+              </div>
+            </div>
+          {% endif %}
+
```
MazOneTwoOne, line 34: Good spot!

fala/apps/adviser/views.py
```diff
@@ -43,33 +42,84 @@ class AccessibilityView(TemplateView):
...
+            self._region = self._form.region
+            if self._region == Region.ENGLAND_OR_WALES:
+                self._data = self._form.search()
+            else:
+                self._data = {}
+        else:
+            self._data = {}
         return super().get(self, request, *args, **kwargs)
 
     # Errors from LaaLaaAPI manifest as _data being falsey, so we bounce back to search page
```
exonian, line 55: I think this comment is now out of date as you now use `self._form.is_valid()` and `self._region` instead, which is a really nice change

fala/apps/adviser/views.py
```diff
@@ -43,33 +42,84 @@ class AccessibilityView(TemplateView):
...
+                    {
+                        "form": self._form,
+                        "data": self._data,
+                        "pages": pages,
+                        "params": params,
+                        "categories": categories,
+                    }
+                )
+            else:
+                region_data = REGION_TO_LINK[self._region]
```
exonian, line 91: It makes me a little uneasy that, while we know that we've written the `get()` method to ensure we don't get here without `self._region`, and that it must be in `REGION_TO_LINK`, this is ultimately pretty brittle and prone to error. I'd like some error handling/fallback behaviour around this, I guess

fala/apps/adviser/views.py
```diff
@@ -43,33 +42,84 @@ class AccessibilityView(TemplateView):
...
+                        "link": region_data["link"],
+                        "region": region_data["region"],
+                    }
+                )
+        else:
+            context.update({"form": self._form, "data": self._data})
         return context
+
+
+REGION_TO_LINK = {
```
exonian, line 104: Weird place for this

fala/apps/adviser/forms.py
```diff
@@ -1,10 +1,12 @@
 # coding=utf-8
+from enum import Enum
+
 from django import forms
 from django.utils.translation import gettext_lazy as _
 from django.utils.safestring import mark_safe
 from django.conf import settings
 
-import laalaa.api as laalaa
+import fala.apps.laalaa.api as laalaa
```
exonian, line 9: It's back :D

fala/apps/adviser/views.py
```diff
@@ -43,33 +42,84 @@ class AccessibilityView(TemplateView):
 class SearchView(ListView):
     def get(self, request, *args, **kwargs):
         self._form = AdviserSearchForm(data=request.GET or None)
-        self._data = self._form.search()
+        if self._form.is_valid():
+            self._region = self._form.region
```
exonian, line 46: I don't remember making so much stuff like this private and it looks strange to my eye, but I could well be wrong. Worth looking at some examples just to check which seems to be more idiomatic.

fala/templates/adviser/other_region.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'base.html' %}
+
+{% block content %}
+  <div class="govuk-grid-column-full">
+    <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
+    <h1>The postcode {{ postcode }} is in {{ region }}</h1>
+    <p class="govuk-body">
+      Find out about <a class="govuk-link" href="{{ link }}">Legal Aid in {{ region }}</a>
+      or try a different search
+    </p>
```
psweeting1, line 10: Where is the change search button? This button when clicked will take the user back to the search page.

fala/apps/adviser/views.py
psweeting1: I almost feel this should be a JSON file on it's own as a resource that we import in and use.

fala/templates/adviser/search.html
psweeting1: This is inside the grey box, but I think it'll be better outside of it. We'll need to check with someone to approve that as there is no design for us in figma for the errors :(

fala/apps/adviser/forms.py
```diff
@@ -16,6 +18,54 @@
     ("Solicitor", "Solicitor"),
 ]
 
+Region = Enum(
```
psweeting1, line 21: Just thinking, can we break these up into their own Python files and import there usage instead of all being in the `forms.py` file.

fala/templates/adviser/other_region.html
starswan: Its on a different ticket

fala/apps/adviser/views.py
starswan: Moved into the view scope

fala/apps/adviser/views.py
starswan: I've deliberately not done this so an exception happens and we get a test failure

starswan: We have come up with a structure that exposes this in a much nicer way using the state pattern

fala/apps/adviser/views.py
starswan: ...and now into the 'OtherJurisdiction' state

fala/apps/adviser/forms.py
starswan: We have moved this to it's own file as it was used by both views and forms

exonian (commented):
I really dislike this "state" pattern.

It turns the views inside out in a weird way (I would expect similar views to extend the same parent and therefore override its methods where needed, but because we have one view here in order to have the convenience of code separation between the ways it works we have a weird sub-view thing living inside our actual view)

If the problem with the old one was that two or three methods each had the same "if" conditional within them, I think the price of this redesign is very not worth it.

In my opinion, the only issues the other design had were that the constants could have been hoisted out. As I'd mentioned, I thought the code structure itself really nicely represented the business logic, and I think that's diminished now. I also think it's just very surprising and if I'm honest it feels like it's trying to be too smart for the problem

fala/apps/adviser/search_view.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def get_queryset(self):
+        return []
+
+    @property
+    def template_name(self):
+        return "other_region.html"
+
+    def get_context_data(self):
+        region_data = self.REGION_TO_LINK[self._region]
```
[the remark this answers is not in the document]
starswan: I guess my question is 'what should happen in this case?' It's actually impossible to trigger at the moment (as this code is only invoked for crown dependencies and Ni/Scotland) and I don't think the set of enums is likely to change?

fala/apps/adviser/views.py
```diff
@@ -37,3 +38,114 @@ def get(self, request, *args, **kwargs):
...
+
+        def get_queryset(self):
+            return []
+
+        @property
+        def template_name(self):
+            return "other_region.html"
+
+        def get_context_data(self):
+            region_data = self.REGION_TO_LINK[self._region]
```
exonian, line 122: OK between the consolidation of related -- and interdependent -- logic into states and me reading through the form class and the view code to understand that we can guarantee that `self._region` exists in `self.REGION_TO_LINK`, I'm now no longer worried about this causing an exception.
I don't super love it as it's far from self-evident why it's not as risky as it looks, but it's good to go.

exonian (approved):
You've won me over on this one 🎉 Very nice work

fala/apps/adviser/search_view.py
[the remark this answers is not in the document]
exonian: It wasn't at all clear without a solid study that self._region was guaranteed to have been set to a value coming from the enum, so I thought values coming straight from form input were being looked up in here, or that the value could be empty or something, but that's not the case.
There's a risk if we add to the enum and forget to add it to the REGION_TO_LINK, but I'm not worried about that in the slightest .

fala/templates/adviser/other_region.html
psweeting1: No, that's not what I'm talking about. There is a GDS green button that allows the user to click and change the search. Look at the design on the Jira ticket. 

The ticket right at the top under the "When" says: `When a user selects change search, they are redirected back to the search page (Landing page).`

fala/apps/adviser/tests/test_search_view_function.py
```diff
@@ -23,25 +23,30 @@ class ResultsPageWithBothOrgAndPostcodeTest(SimpleTestCase):
...
-            '<span class="results-header"> <span class="govuk-!-font-weight-bold">22 results</span> in order of closeness to'
-            + '<strong class="notranslate" translate="no">pe30</strong>'
-            + "matching <strong>bu</strong>."
-            + "</span>"
-        )
-        self.assertContains(
-            response,
-            expected,
-            html=True,
-        )
```
starswan, line 40: removing this feels like a backward step - the intention of the test has changed from finding a specific state so asserting just 'presence'

fala/apps/adviser/tests/test_search_view_function.py
```diff
@@ -61,14 +66,29 @@ class ResultsPageWithJustPostcodeTest(SimpleTestCase):
...
 
-    def test_category_search_heading(self):
+    def test_category_search_heading_postcode_only(self):
         response = self.client.get(self.url, self.data)
-        self.assertContains(
-            response,
-            '<span class="results-header"><span class="govuk-!-font-weight-bold">350 results</span> in order of closeness to'
-            + '<strong class="notranslate" translate="no">PE30</strong> . </span>',
-            html=True,
-        )
```
starswan, line 71: see above comment

MazOneTwoOne (approved):
Happy to approve. 

I had the one question about the 'Error Summary' not showing as expected (but I believe this will be picked up in another ticket??) - my question could also be a "red herring".

fala/templates/adviser/search.html
```diff
@@ -10,6 +10,19 @@
 {% block content %}
   <div class="govuk-grid-column-two-thirds">
       <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
+      {% if (form.postcode and form.postcode.errors) or (form.name and form.name.errors) %}
```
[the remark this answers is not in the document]
starswan: There appears to be a weird issue here - if you type in ZZ1 you get the error summary box, but if you type in nothing then just the error shows up w/o the summary

PR #243, opened 2024-04-24
EL-1436 google map link on results page

starswan:

Add google map link on results page

## Checklist

### DIFF
--- a/fala/apps/adviser/templatetags/adviser_extras.py
+++ b/fala/apps/adviser/templatetags/adviser_extras.py
@@ -34,4 +34,9 @@ def to_fala_page_url(context, value, url_path):
 
 @library.filter
 def google_map_params(location):
-    return {"api": 1, "query": location["address"] + " " + location["postcode"]}
+    if "postcode" in location and "address" in location:
+        return {"api": 1, "query": f"{location['address']} {location['postcode']}"}
+    elif "address" in location:
+        return {"api": 1, "query": location["address"]}
+    else:
+        return {"api": 1, "query": location["postcode"]}
--- a/fala/apps/adviser/templatetags/adviser_extras.py
+++ b/fala/apps/adviser/templatetags/adviser_extras.py
@@ -30,3 +30,8 @@ def query_to_dict(_context, value, prop=None):
 @pass_context
 def to_fala_page_url(context, value, url_path):
     return "%s?%s" % (url_path, urlparse(value).query)
+
+
+@library.filter
+def google_map_params(location):
+    return {"api": 1, "query": location["address"] + " " + location["postcode"]}
--- a/fala/apps/adviser/tests/test_adviser_extras.py
+++ b/fala/apps/adviser/tests/test_adviser_extras.py
@@ -15,3 +15,8 @@ class QueryToDictTest(unittest.TestCase):
     def test_returns_an_empty_list_when_a_prop_is_given_but_does_not_exist(self):
         result = adviser_extras.query_to_dict("irrelevant", "http://localhost:8000/?postcode=SW1A&page=3", "potato")
         self.assertEqual([], result)
+
+    def test_google_map_params(self):
+        location = {"address": "The Core, County Way Barnsley", "postcode": "S70 2JW"}
+        result = adviser_extras.google_map_params(location)
+        self.assertEqual(result, {"api": 1, "query": "The Core, County Way Barnsley S70 2JW"})
--- a/fala/apps/adviser/tests/test_adviser_extras.py
+++ b/fala/apps/adviser/tests/test_adviser_extras.py
@@ -20,3 +20,13 @@ class QueryToDictTest(unittest.TestCase):
         location = {"address": "The Core, County Way Barnsley", "postcode": "S70 2JW"}
         result = adviser_extras.google_map_params(location)
         self.assertEqual(result, {"api": 1, "query": "The Core, County Way Barnsley S70 2JW"})
+
+    def test_map_params_without_postcode(self):
+        location = {"address": "The Core, County Way Barnsley"}
+        result = adviser_extras.google_map_params(location)
+        self.assertEqual(result, {"api": 1, "query": "The Core, County Way Barnsley"})
+
+    def test_map_params_without_address(self):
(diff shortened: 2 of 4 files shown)


fala/apps/adviser/templatetags/adviser_extras.py
```diff
@@ -30,3 +30,8 @@ def query_to_dict(_context, value, prop=None):
 @pass_context
 def to_fala_page_url(context, value, url_path):
     return "%s?%s" % (url_path, urlparse(value).query)
+
+
+@library.filter
+def google_map_params(location):
```
psweeting1, line 36:
```suggestion
def google_map_params(location):
    return {"api": 1, "query": f"{location['address']} {location['postcode']}"}
```

fala/templates/adviser/results.html
```diff
@@ -57,14 +57,17 @@ <h2 class="govuk-body-l govuk-!-font-weight-bold">{{ item.organisation.name }}</
             {% endif %}
             {% if item.categories|length %}
             <div class="categories">
-              <h3 class="govuk-heading-s govuk-!-margin-bottom-2">{{ _('Categories') }}</h3>
+              <h3 class="govuk-heading-s govuk-!-margin-bottom-2">{{ _('Categories ') }}</h3>
               <ul class="govuk-list govuk-list--bullet">
                 {% for cat in item.categories %}
-                  <li>{{ cat }}</li>
+                  <li >{{ cat }}</li>
```
psweeting1, line 63:
```suggestion
                  <li>{{ cat }}</li>
```

fala/templates/adviser/results.html
```diff
@@ -57,14 +57,17 @@ <h2 class="govuk-body-l govuk-!-font-weight-bold">{{ item.organisation.name }}</
...
                 {% endfor %}
               </ul>
             </div>
             {% endif %}
+            <a class="govuk-link" class="url" target="_blank" rel="noopener" href="https://www.google.com/maps/search/?{{ item.location|google_map_params|urlencode }}">
```
psweeting1, line 68: **Optional** I'd want your take on this. But I feel `https://www.google.com/maps/search/?` shouldn't be in a template, but in a variable accessible in the template.

fala/apps/adviser/tests/test_adviser_extras.py
```diff
@@ -15,3 +15,8 @@ def test_returns_the_value_of_a_query_parameter_when_a_prop_is_given(self):
     def test_returns_an_empty_list_when_a_prop_is_given_but_does_not_exist(self):
         result = adviser_extras.query_to_dict("irrelevant", "http://localhost:8000/?postcode=SW1A&page=3", "potato")
         self.assertEqual([], result)
+
+    def test_google_map_params(self):
```
psweeting1, line 19: I'd personally like to see more tests for different case scenarios instead of just a happy one. 

- Test without address
- Test without postcode
- Test that the function handles None values for the location parameter gracefully. (or Unexpected values) how does it handle this?

psweeting1 (changes requested):

fala/apps/adviser/tests/test_adviser_extras.py
starswan: It doesn't, and I couldn't find any reason for the code to be more complicated than it currently is.

fala/apps/adviser/templatetags/adviser_extras.py
starswan: I'm not sure that there is any merit in your solution over mine?

fala/templates/adviser/results.html
starswan: This felt the simplest way of doing it. The template already knows this is a search (the filter function returns parameters for a search) so adding complexity didn't feel worth it

fala/apps/adviser/templatetags/adviser_extras.py
starswan: This feels more complicated than my version

psweeting1: Yes it's tiny use here, I agree, but I'd always recommend using f-string. f-strings support the full power of Python expressions, allowing you to embed variables, function calls, and other expressions directly within the string. 
This is normal in Python to use f-string where possible over using concatenation strings. I'd recommend the use of f-string anytime over concatenation strings where possible. 

Sorry I personally disagree on the readability, that doesn't mean I'm write of course and that comes down to personal preference on some cases but consider the following. f-strings make the code more readable by allowing you to embed expressions directly within the string, surrounded by curly braces ({}). This eliminates the need for complex string concatenation and helps improves code clarity. f-strings also help reduce boilerplate code by providing a concise syntax for string formatting. 

Whenever I see `+ "" +`  in code, I cry a little bit on the inside :(

fala/apps/adviser/tests/test_adviser_extras.py
psweeting1: But for defence purposes, we shouldn't always assume that the data is always going to be fine right though surely?

fala/apps/adviser/templatetags/adviser_extras.py
exonian: This looks to have been resolved now?

fwiw I definitely agree that string concatenation in Python code sets off the little alarm bells for me. I would always go with a `string.format()` or, if I were less old-fashioned, the newer and better f-strings :D

fala/apps/adviser/tests/test_adviser_extras.py
exonian: I think Pat's right here. At the very least, I don't want future devs having to come here and work out they need to write extra tests when they want to change laalaa because we haven't been a little bit defensive

fala/templates/adviser/results.html
starswan: I thought about google_maps_link but then wanted the URL encoding out of that method, so this seemed to be a good compromise

fala/templates/adviser/results.html
starswan: thanks - I have no idea how that artifact got in ther

psweeting1 (approved):
LGTM

PR #248, opened 2024-04-29
EL-1435: Display params on results page and allow return to search

psweeting1:

This pull request implements the grey 'change search' on the results.html. The box shows the user their current search parameters and also have a button that allows them to go back to the search.html and perform a new search.

When they change their search, their previous search params are populated on the search page for them to edit.

When they do not add postcode or org, this param will not render in the grey box on results.html

To translate the categories on the results.html to nice strings, we have had to use the code which gets them from the LAALAA api. ~~My only concern is we will be hitting this API a lot more to get the categories.~~

~~I couldn't see a simple way to just store whatever is returned as the categories from the API (was thinking something like Ruby's `@variable ||=` but couldn't find a good python equivalent).~~

~~We could implement caching but that seemed like a fairly big-ish change so I haven't done it for now. Maybe a follow up ticket if we think we're hitting the API too much? Maybe it is already caching?~~

Update - the categories are being loaded when the application starts so they are saved/cached already. No further caching work needed!

## Checklist

### DIFF
--- a/fala/apps/adviser/tests/test_search_view_function.py
+++ b/fala/apps/adviser/tests/test_search_view_function.py
@@ -39,6 +39,12 @@ class ResultsPageWithBothOrgAndPostcodeTest(SimpleTestCase):
             html=True,
         )
 
+    def test_search_parameters_box_is_visible(self):
+        response = self.client.get(self.url, self.data)
+        soup = bs4.BeautifulSoup(response.content)
+        search_params_box = soup.find("div", class_="laa-fala__grey-box")
+        self.assertIsNotNone(search_params_box)
+
     def test_search_returns_results_list(self):
         response = self.client.get(self.url, self.data)
         soup = bs4.BeautifulSoup(response.content, "html.parser")
@@ -69,6 +75,30 @@ class ResultsPageWithJustPostcodeTest(SimpleTestCase):
             + '<strong class="notranslate" translate="no">PE30</strong> . </span>',
             html=True,
         )
+
+    def test_search_parameters_box_contains_only_postcode_and_categories(self):
+        response = self.client.get(self.url, self.data)
+        soup = bs4.BeautifulSoup(response.content)
+        search_params_box = soup.find("div", class_="laa-fala__grey-box")
+        # replace the spaces in the HTML for ease of comparison
+        content = search_params_box.text.replace("\n", "")
+        self.assertEqual(content, "Postcode: PE30Categories: Debt Change search")
+
+
+@override_settings(FEATURE_FLAG_NO_MAP=True)
+class ResultsPageWithJustOrgTest(SimpleTestCase):
+    client = Client()
+    url = reverse("search")
+
+    data = {"name": "foo", "categories": ["deb", "edu"]}
+
+    def test_search_parameters_box_contains_only_organisation_and_categories(self):
+        response = self.client.get(self.url, self.data)
+        soup = bs4.BeautifulSoup(response.content)
+        search_params_box = soup.find("div", class_="laa-fala__grey-box")
+        # replace the spaces in the HTML for ease of comparison
+        content = search_params_box.text.replace("\n", "")
+        self.assertEqual(content, "Organisation: fooCategories: Debt, Education Change search")
 
 
 @override_settings(FEATURE_FLAG_NO_MAP=True)
--- a/fala/templates/adviser/results.html
+++ b/fala/templates/adviser/results.html
@@ -3,7 +3,7 @@
 {% block content %}
   <div class="govuk-grid-column-full">
     <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
-    <h1>Search results</h1>
+    <h1 class="govuk-heading-xl">Search results</h1>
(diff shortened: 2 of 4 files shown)


fala/apps/adviser/tests/test_search_view_function.py
```diff
@@ -39,6 +39,10 @@ def test_category_search_heading_closeness_and_matching(self):
             html=True,
         )
 
+    def test_search_parameters_box_is_visible(self):
+        response = self.client.get(self.url, self.data)
+        self.assertContains(response, '<div class="laa-fala__grey-box')
```
psweeting1, line 44: I think BS4 would be better here:

```
soup = bs4.BeautifulSoup(response.content)
search_params_box = soup.find("div", class_="laa-fala__grey-box")
```

And then assert it's not None. This means that it found the HTML and it's content is not empty.

If you run into trouble with that and still feel assertContains is fine, then please add the closing `>` To the HTML response check.

fala/apps/adviser/tests/test_search_view_function.py
```diff
@@ -70,6 +74,13 @@ def test_category_search_heading(self):
             html=True,
         )
 
+    def test_search_parameters_box_contains_only_postcode_and_categories(self):
```
psweeting1, line 77: Great but a test for any categories would also be great to see to ensure that is correctly printing them into the DOM.

fala/apps/adviser/tests/test_search_view_function.py
```diff
@@ -70,6 +74,13 @@ def test_category_search_heading(self):
...
+        response = self.client.get(self.url, self.data)
+        soup = bs4.BeautifulSoup(response.content)
+        search_params_box = soup.find("div", class_="laa-fala__grey-box")
+        content = search_params_box.text.replace("\n", "")
```
psweeting1, line 81: Can you just add a comment to this text to say why you're removing spaces, hence why the string check below is slightly clumped together in the assertEquals. I know why, because of how the HTML is displaying the Strings, but someone else in the future may get confused.

fala/templates/adviser/results.html
```diff
@@ -4,6 +4,31 @@
   <div class="govuk-grid-column-full">
     <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
     <h1>Search Results</h1>
+    <div class="laa-fala__grey-box govuk-!-margin-bottom-6">
```
psweeting1, line 7: This needs a role attribute added. `role="list"`

fala/templates/adviser/results.html
```diff
@@ -4,6 +4,31 @@
...
+      <ul class="govuk-list">
+        {% if form.postcode.value() %}
+          <li class="govuk-body notranslate" translate="no">Postcode: {{ form.postcode.value() }}</li>
```
psweeting1, line 10: role attribute needed for each `<li>`. `role="listitem"`

exonian:
I agree with your assessment that this _should_ be cached in some way, but that we can tolerate releasing this without caching and that merging and delivering is preferable, with that being done in a future ticket.

It should be quite straightforward to do the approach you're suggesting, the trick is just deciding where to put the variable to ensure it's available where we want it. And then once we know that, we can decide how to store it. While we could do the caching ourselves with variable assignment and an `if` statement, and this would certainly save us configuring Django's cache, I think it would be preferable to bite the bullet and get Django's way set up.

willc-work (approved):

exonian:
> I agree with your assessment that this _should_ be cached in some way, but that we can tolerate releasing this without caching and that merging and delivering is preferable, with that being done in a future ticket.
> 
> It should be quite straightforward to do the approach you're suggesting, the trick is just deciding where to put the variable to ensure it's available where we want it. And then once we know that, we can decide how to store it. While we could do the caching ourselves with variable assignment and an `if` statement, and this would certainly save us configuring Django's cache, I think it would be preferable to bite the bullet and get Django's way set up.

That'll teach me to not actually go look at [the API consumer in cla_common](https://github.com/ministryofjustice/cla_common/blob/master/cla_common/laalaa.py#L16-L17), which does of course has exactly this built in.

HettieS:
> > I agree with your assessment that this _should_ be cached in some way, but that we can tolerate releasing this without caching and that merging and delivering is preferable, with that being done in a future ticket.
> > It should be quite straightforward to do the approach you're suggesting, the trick is just deciding where to put the variable to ensure it's available where we want it. And then once we know that, we can decide how to store it. While we could do the caching ourselves with variable assignment and an `if` statement, and this would certainly save us configuring Django's cache, I think it would be preferable to bite the bullet and get Django's way set up.
> 
> That'll teach me to not actually go look at [the API consumer in cla_common](https://github.com/ministryofjustice/cla_common/blob/master/cla_common/laalaa.py#L16-L17), which does of course has exactly this built in.

Nice! I hadn't actually looked at that and it's good to see the implementation there. It was actually something else I talked about with @starswan, that the variables which hold the category data are actually defined as constants in `laalaa/api.py`. But I'm guessing it's the line you found that is actually loading and then storing them.

PR #250, opened 2024-04-29
EL:1473: Replace local.py with .env

MazOneTwoOne:

- replaced `local.py` with `.env`
- updated readme
- updated shell script
- updated requirements 

- please also make sure you can, run this locally in both a docker container and using virtual python env

## Checklist


exonian (commented):
I would suggest a comment in the readme about the Fun we discovered this morning about the unexpected behaviour when changing the contents of the .env file, assuming no way around that was found after I dropped off

MazOneTwoOne:
> I would suggest a comment in the readme about the Fun we discovered this morning about the unexpected behaviour when changing the contents of the .env file, assuming no way around that was found after I dropped off

This is now resolved by this line and added a comment:

```
# With override set to True the value of the variable in `.env` is loaded first https://pypi.org/project/python-dotenv/#variable-expansion
load_dotenv(override=True)
```

HettieS:
Did you check the docker config still works? It would be good to highlight that briefly on the PR in the relevant section

MazOneTwoOne:
> Did you check the docker config still works? It would be good to highlight that briefly on the PR in the relevant section

 👍 Yep. Checked that we can still run a local docker container. No need to update of `readme.md` as none of those steps have changed. (I had a slight issue with CSS style's but not docker related).

I therefore asked Stephen to pull down the branch on his machine and he was able to run a docker container as expected.

HettieS:
> > Did you check the docker config still works? It would be good to highlight that briefly on the PR in the relevant section
> 
> 👍 Yep. Checked that we can still run a local docker container. No need to update of `readme.md` as none of those steps have changed. (I had a slight issue with CSS style's but not docker related).
> 
> I therefore asked Stephen to pull down the branch on his machine and he was able to run a docker container as expected.

Nice one. I didn't mean to update the README, just that it might be best practice to update the PR info to highlight that to the reviewer if possible, as it's not obvious we've thought about checking Docker from the code. Just PR housekeeping for next time! Good work on this!

psweeting1 (approved):
LGTM Nice work.

psweeting1 (approved):
LGTM

PR #254, opened 2024-05-10
Update fala to deploy production from main after branch rename

starswan:

Re-enable production deploys after master branch rename

## Checklist


exonian:
I think we also need to fix up the image tagging. At a glance, it looks like it'll tag them `main` but other things are still looking for a `master` tag image

PR #259, opened 2024-05-16
EL-1467 Add Research banner

starswan:

Add research banner

None

## Checklist

### DIFF
--- a/fala/settings/base.py
+++ b/fala/settings/base.py
@@ -245,4 +245,4 @@ GOOGLE_MAPS_API_KEY = os.environ.get("GOOGLE_MAPS_API_KEY", "")
 
 CHECK_LEGAL_AID_URL = "https://www.gov.uk/check-legal-aid"
 
-FEATURE_FLAG_SURVEY_MONKEY = os.environ.get("FEATURE_FLAG_SURVEY_MONKEY")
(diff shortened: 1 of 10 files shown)


fala/settings/base.py
```diff
@@ -235,3 +235,5 @@ def root(*x):
 GOOGLE_MAPS_API_KEY = os.environ.get("GOOGLE_MAPS_API_KEY", "")
 
 CHECK_LEGAL_AID_URL = "https://www.gov.uk/check-legal-aid"
+
+FEATURE_FLAG_SURVEY_MONKEY = os.environ.get("FEATURE_FLAG_SURVEY_MONKEY")
```
HettieS, line 239: do we need a fall back of `""` or `False` as with other env variables? The FEATURE_FLAG_NO_MAP is set up slightly differently so wonder if we need that implementation here too.

HettieS: Plus we are not specifying the value of the flag in dev or prod in this PR. I think that might be fine but I think I'd prefer explicit defaults.

HettieS (commented):
Looking good, a question around the feature flag set up!

HettieS (approved):
LGMT

PR #263, opened 2024-05-24
EL-1508 add new feedback table

willc-work:

[Jira Ticket](https://dsdmoj.atlassian.net/browse/EL-1508)
Adds a feedback table to the database

Creates a k8s job to run the migration via a script
When this has been merged with main I will need to apply the migrate-db-job.yml which will run the script
There are options to automate this via circle, but there are better options available to us once we are on helm (EL-1463) So I will write a ticket to undertake that work once both these current tickets are merged.

Intentionally left blank.

## Checklist


.env.example
```diff
@@ -7,6 +7,7 @@ LAALAA_API_HOST = "https://laa-legal-adviser-api-staging.apps.live-1.cloud-platf
 ENVIRONMENT = "dev"
 FEATURE_FLAG_NO_MAP = False
 
+DB_NAME=fala_db
```
exonian, line 10: This is already [the default in our settings](https://github.com/ministryofjustice/fala/pull/263/files#diff-75397501301f0e287528185a96d3452d3200007bd6c068d77cdd489e1a18c987R42), so I don't know if it's valuable to declare it here too

fala/apps/adviser/migrations/0002_alter_feedback_id.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 5.0.6 on 2024-05-24 12:47
```
exonian, line 1: While having any number of migrations is fine and once we're under way there's no need to be thrifty with them (as stated explicitly in the docs themselves) I would recommend it's good practice at this point to combine this and the above migration. If nothing else, it's good exercise.

Your options are likely:
- delete both migrations and run `makemigrations` again
- use `squashmigrations` ([docs here](https://docs.djangoproject.com/en/5.0/topics/migrations/#squashing-migrations))
- manually edit the first migration

Although each of these, possibly apart from the second, probably involves blowing away your local db first as it keeps track of what migrations have been applied and deleting migrations would impact that. If you can migrate back to _before_ the initial migration (I know back in the day you couldn't but that may have changed) then you could do that, then delete the migrations, then make new ones, then migrate.

So yeah, maybe it's not worth it at all 😂

fala/apps/adviser/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from django.utils import timezone
+
+
+class Feedback(models.Model):
+    satisfied = models.BooleanField(null=True, blank=True)
+    comment = models.TextField(null=True, blank=True)
```
exonian, line 7: Null is not recommended on Char or Text fields (or their derivatives) in Django. Basically you don't want empty strings _and_ nulls in a column, and [Django's convention](https://docs.djangoproject.com/en/5.0/ref/models/fields/#null) is to disallow null and therefore use empty strings

exonian (commented):
You are correct that running `migrate` will create, destroy and modify db tables as needed, but will not create the `fala_db` itself. So once we have a db in a given environment (and specifically a db not just a postgres container), running migrations will be all we need to maintain it.

fala/apps/adviser/migrations/0002_alter_feedback_id.py
willc-work: Thanks, I understand now what happened, the original migration failed and so I added the BigAutoField and re-ran the migration, thinking I would get one successful migration. I will do as you suggest and blow it away and then rerun them as a single migration.

docs/virtual-env.md
```diff
@@ -92,6 +92,7 @@ node --version
 
 - You can use `psql -l` to list available databases, then exit. This will show the credentials to use for `DB_USER` in the 'Owner' column. Update you `.env` file with the following:
 
+       DB_NAME=fala_db
```
exonian, line 95: For the same reasoning we removed it from the example env file I think it doesn't need to be here in the readme, but it's not a big deal.

fala/apps/adviser/models.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class SatisfactionFeedback(models.Model):
+    satisfied = models.BooleanField(null=True, blank=True)
+    comment = models.TextField(blank=True)
+    created_at = models.DateTimeField(default=timezone.now)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        return f"Feedback {self.id}"
```
exonian, line 12: The linter does not like you having removed the newline char

exonian (approved):
Linting need appeasing, and a non-blocking suggestion you can action or not

exonian (approved):

kubernetes_deploy/staging/create-db-job.yml
```diff
@@ -0,0 +1,35 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: create-db-run-migrations
+spec:
+  template:
+    spec:
+      containers:
+      - name: create-and-migrate
```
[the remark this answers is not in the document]
willc-work: fixed in subsequent commit

PR #267, opened 2024-06-03
El 1490 remove s3 dependency

starswan:

Remove the dependency on S3. Moves static files into file system alongside code

This change requires a fix to the end2end tests - they were broken as they were being run against a container without CSS styling, so the GDS styles were not hiding the checkboxes as they do in reality

## Checklist


kubernetes_deploy/staging/deployment.yml
```diff
@@ -60,14 +60,7 @@ spec:
...
-        - name: STATIC_FILES_BACKEND
-          value: s3
-        - name: AWS_STORAGE_BUCKET_NAME
-          valueFrom:
-            secretKeyRef:
-              name: s3
-              key: bucket_name
         - name: FEATURE_FLAG_NO_MAP
-          value: "true"
+          value: "false"
```
psweeting1, line 64: I think if we merged this as `false`, anyone working on the new design of FALA will not see the new design page coming through on staging. So they'll turn it back on again in a different commit probably and merge it. 

I think this should be left `true`

PR #271, opened 2024-06-06
Wrapped urlpattern to search page in feature flag condition

psweeting1:

Required.

Intentionally left blank.

## Checklist


fala/apps/adviser/tests/test_crown_dependencies_view.py
```diff
@@ -6,7 +6,10 @@
 @override_settings(FEATURE_FLAG_NO_MAP=True)
 class PostcodeValidationTest(SimpleTestCase):
     client = Client()
-    url = reverse("search")
+
```
exonian, line 9: This no longer seems necessary if we're not varying the urlconf, but do we like it anyway?

exonian (approved):

PR #276, opened 2024-06-13
EL-1536 Scotland region displays results page.

psweeting1:

- Adding in Ariel label for hint text on postcode field.
- Created new Scotland HTML to display if postcode region is Scotland.
- Updated `view.py` to include Scotland Region logic for England or Wales.
- Added new tests and updated existing to be parasitised. 

- Search a Scottish postcode to see alert on results page. 
- Search a none Scottish postcode to ensure all other regions are supported
- Search an England postcode to ensure warning message does not appear.

## Checklist

- Given I am searching on FALA
- And the postcode hint as has aria label.
- And I entered a Scotland postcode
- When I enter a Scotland postcode
- I see results for England and Wales
- And a message telling me that the results only cover these two countries
- [EL-1536](https://dsdmoj.atlassian.net/browse/EL-1536)

[EL-1536]: https://dsdmoj.atlassian.net/browse/EL-1536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ docs/_build
 env
 env*
 venv
+.venv
 
 # intellij
 .idea/
@@ -54,12 +55,12 @@ bower_components/
 
 .sass-cache/
 
-venv
-
 # Fala
 .env
 fala/assets
 fala/static
+.cache
+.bash_history
 
 # Python version on local machine
 .python-version
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,6 @@ docs/_build
 env
 env*
 venv
-.venv
 
 # intellij
 .idea/
--- a/fala/apps/adviser/tests/test_crown_dependencies_view.py
+++ b/fala/apps/adviser/tests/test_crown_dependencies_view.py
@@ -13,28 +13,29 @@ class PostcodeValidationTest(SimpleTestCase):
         response = self.client.get(self.url, data)
         self.assertContains(response, "The postcode GY1 is in Guernsey")
 
-    def test_invalid_guernsey_postcode(self):
-        data = {"postcode": "GY1"}
+    def test_northern_ireland_postcode(self):
+        data = {"postcode": "BT93 8AD"}
         response = self.client.get(self.url, data)
-        self.assertContains(response, "Legal Aid in Guernsey")
-
-    def test_invalid_scottish_postcode(self):
-        data = {"postcode": "AB"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode AB is in Scotland")
+        self.assertContains(response, "The postcode BT93 8AD is in Northern Ireland")
 
     def test_valid_scottish_postcode(self):
         data = {"postcode": "AB11 5BN"}
         response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode AB11 5BN is in Scotland")
+        self.assertContains(response, "These results cover England and Wales.")
 
     def dont_test_valid_english_postcode(self):
         data = {"postcode": "M2 3WQ"}
         response = self.client.get(self.url, data)
         self.assertContains(response, "in order of closeness to")
 
+    def test_lowercase_postcode_passes(self):
+        # Lower case postcode string
+        data = {"postcode": "im4"}
+        response = self.client.get(self.url, data)
+        self.assertContains(response, "The postcode IM4 is in the Isle of Man")
+
     def test_change_search_button_and_form_is_visible(self):
-        data = {"postcode": "AB11 5BN"}
+        data = {"postcode": "IM4"}
         response = self.client.get(self.url, data)
         soup = bs4.BeautifulSoup(response.content, "html.parser")
         form = soup.find("form", {"action": "/", "method": "get"})
@@ -42,12 +43,6 @@ class PostcodeValidationTest(SimpleTestCase):
         self.assertIsNotNone(form)
         self.assertIsNotNone(button)
         self.assertEqual(button.text.strip(), "Change search")
-
-    def test_lowercase_postcode_passes(self):
-        # Lower case postcode string
-        data = {"postcode": "ab"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode AB is in Scotland")
 
 
 @override_settings(FEATURE_FLAG_NO_MAP=True)
--- a/fala/apps/adviser/tests/test_crown_dependencies_view.py
+++ b/fala/apps/adviser/tests/test_crown_dependencies_view.py
@@ -8,33 +8,30 @@ class PostcodeValidationTest(SimpleTestCase):
     client = Client()
     url = reverse("search")
 
-    def test_guernsey_postcode(self):
-        data = {"postcode": "GY1"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode GY1 is in Guernsey")
+    def test_region_postcodes(self):
+        test_cases = [
+            {"postcode": "GY1", "message": "The postcode GY1 is in Guernsey"},
+            {"postcode": "BT93 8AD", "message": "The postcode BT93 8AD is in Northern Ireland"},
+            {"postcode": "IM4", "message": "The postcode IM4 is in the Isle of Man"},
+            {"postcode": "JE2 3FN", "message": "The postcode JE2 3FN is in Jersey"},
+            # English Postcode
+            {"postcode": "M2 3WQ", "message": "in order of closeness to"},
+            # Scottish Postcode
+            {"postcode": "AB11 5BN", "message": "These results cover England and Wales."},
+            # Lower case Postcode
+            {"postcode": "im4", "message": "The postcode IM4 is in the Isle of Man"},
+            # Invalid post code with no prefix numbers, results are found and search
+            # Region is recognised as Scotland.
+            {"postcode": "AB", "message": "These results cover England and Wales."},
+        ]
 
-    def test_northern_ireland_postcode(self):
-        data = {"postcode": "BT93 8AD"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode BT93 8AD is in Northern Ireland")
+        for case in test_cases:
+            with self.subTest(postcode=case["postcode"]):
+                data = {"postcode": case["postcode"]}
+                response = self.client.get(self.url, data)
+                self.assertContains(response, case["message"])
 
-    def test_valid_scottish_postcode(self):
-        data = {"postcode": "AB11 5BN"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "These results cover England and Wales.")
-
-    def dont_test_valid_english_postcode(self):
-        data = {"postcode": "M2 3WQ"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "in order of closeness to")
-
-    def test_lowercase_postcode_passes(self):
-        # Lower case postcode string
-        data = {"postcode": "im4"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode IM4 is in the Isle of Man")
-
-    def test_change_search_button_and_form_is_visible(self):
+    def test_other_region_form_and_change_search_button_visible(self):
         data = {"postcode": "IM4"}
         response = self.client.get(self.url, data)
         soup = bs4.BeautifulSoup(response.content, "html.parser")
@@ -42,7 +39,10 @@ class PostcodeValidationTest(SimpleTestCase):
         button = soup.find("button", {"type": "submit", "data-module": "govuk-button"})
         self.assertIsNotNone(form)
         self.assertIsNotNone(button)
-        self.assertEqual(button.text.strip(), "Change search")
+
+        change_search_button = soup.find("button", {"id": "otherRegionChangeSearchButton"})
+        self.assertIsNotNone(change_search_button)
+        self.assertEqual(change_search_button.text.strip(), "Change search")
 
 
 @override_settings(FEATURE_FLAG_NO_MAP=True)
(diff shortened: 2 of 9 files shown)


fala/apps/adviser/tests/test_crown_dependencies_view.py
```diff
@@ -8,33 +8,28 @@ class PostcodeValidationTest(SimpleTestCase):
...
-        data = {"postcode": "GY1"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "The postcode GY1 is in Guernsey")
-
-    def test_invalid_guernsey_postcode(self):
-        data = {"postcode": "GY1"}
-        response = self.client.get(self.url, data)
-        self.assertContains(response, "Legal Aid in Guernsey")
-
-    def test_invalid_scottish_postcode(self):
```
MazOneTwoOne, line 21: question - do we want to keep some of the invalid test cases i.e. 

- Invalid Guernsey postcode
- Invalid Scotland postcode

fala/apps/adviser/tests/test_crown_dependencies_view.py
```diff
@@ -43,12 +38,6 @@ def test_change_search_button_and_form_is_visible(self):
         self.assertIsNotNone(button)
         self.assertEqual(button.text.strip(), "Change search")
```
MazOneTwoOne, line 39: question - would it be better to search for the id of the 'Change search' button? 

The results pages both have   'Change search' buttons but helpfully now have different id's: `otherRegionChangeSearchButton` & `changeSearchButton`

fala/templates/adviser/scotland_warning.html
```diff
@@ -0,0 +1,11 @@
+<div class="govuk-grid-column-full">
```
MazOneTwoOne, line 1: This should be `class="govuk-grid-column-two-thirds"` in order to get the text to wrap.

fala/templates/adviser/scotland_warning.html
```diff
@@ -0,0 +1,11 @@
...
+  <div class="govuk-warning-text" id="legal_aid_in_scotland_warning_text">
+    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+    <strong class="govuk-warning-text__text">
+      <span class="govuk-warning-text__assistive">Warning </span>
+      <p>These results cover England and Wales. Find out about
+        <a class="govuk-link" href="https://www.mygov.scot/legal-aid/">Legal Aid in Scotland.</a>
```
MazOneTwoOne, line 7: question - this question is more for the content designers, but should this link now open on a new tab? Otherwise you see the results and may click away from the screen.

fala/templates/macros/forms.html
```diff
@@ -54,9 +54,9 @@
...
-            <input type="text" name="{{ field.name }}" value="{{ field.data or '' }}" class="govuk-input govuk-input--error govuk-!-width-one-third {{ field_attrs.class or '' }}" maxlength="{{ kwargs.max_length or '30' }}" id="id_{{ field.name }}">
+            <input type="text" name="{{ field.name }}" value="{{ field.data or '' }}" class="govuk-input govuk-input--error govuk-!-width-one-third {{ field_attrs.class or '' }}" maxlength="{{ kwargs.max_length or '30' }}" id="id_{{ field.name }}" aria-describedby="field-description-postcode">
           {% else %}
-            <input type="text" name="{{ field.name }}" value="{{ field.data or '' }}" class="govuk-input govuk-!-width-one-third {{ field_attrs.class or '' }}" maxlength="{{ kwargs.max_length or '30' }}" id="id_{{ field.name }}">
+            <input type="text" name="{{ field.name }}" value="{{ field.data or '' }}" class="govuk-input govuk-!-width-one-third {{ field_attrs.class or '' }}" maxlength="{{ kwargs.max_length or '30' }}" id="id_{{ field.name }}" {% if field.name == 'postcode' %}aria-describedby="field-description-postcode"{% endif %}>
```
MazOneTwoOne, line 59: question - do we need to add a comment to say that we should remove this 'conditional if statement' once FALA refresh has been launched?

MazOneTwoOne (commented):
Just some questions, additions and requests but overall looks good. 😊

fala/templates/adviser/scotland_warning.html
psweeting1: A good question, I can ask.

Just my thoughts, on the current region support page, we do not open in a new tab and the design provided does not have the hint text `(Opens in a new tab)` appended to it. So the ticket and design implies that open in a new tab was considered already, but no harm in asking.

fala/templates/macros/forms.html
psweeting1: I can add a comment in regards to why this is appended to the this input sure.

In regards to Fala's redesign, this will still be needed unless we gut and refactor everything in terms of generating forms in FALA which is a big piece of work that isn't in the pipelines from what we've planned in terms of these forms.

fala/apps/adviser/tests/test_crown_dependencies_view.py
psweeting1: Yeah, I can add one test in there.

Those current tests proved nothing unfortunately. It's the same postcode being used, it's just checking a different string on the page. 

They are also not invalid postcodes, they are valid. 

Invalid postcodes are handled by the search input field which already have tests to check for an invalid postcode. 

I will test however a bad postcode against the views to make sure it handles it fine.

fala/templates/macros/forms.html
MazOneTwoOne: Fair enough, I hadn't realised we weren't going to 'gut out' the old design once we released the new design.

fala/apps/adviser/tests/test_crown_dependencies_view.py
psweeting1: Sure, I can be cheeky and add that in.

fala/apps/adviser/tests/test_crown_dependencies_view.py
MazOneTwoOne: Thank you. 

I noticed a few inconsistencies when entering part of a postcode, hence the comment. e.g:

- 'BT' - does a search
- 'BT [and number] - goes to NI page.

- 'IM' - should invalid postcode error message
- 'IM [and number] - goes to Isles of Man page.

fala/templates/adviser/scotland_warning.html
psweeting1: Oooof! Good catch.

fala/templates/macros/forms.html
psweeting1: This is still part of the new design. It's just an input field.

fala/apps/adviser/tests/test_crown_dependencies_view.py
psweeting1: Yeah, the postcode check still need fine tuning and there is more work to be done on it. For example TD15 is not a Scotland postcode even when we say it is.

It's out of scope of this ticket to go and fix this and the above. However bug tickets and feature tickets should be created from this. 

I've found we handled postcode not found bad requests poorly, which is something we'll also need to fix later.

fala/apps/adviser/tests/test_crown_dependencies_view.py
```diff
@@ -8,46 +8,42 @@ class PostcodeValidationTest(SimpleTestCase):
...
+            {"postcode": "BT93 8AD", "message": "The postcode BT93 8AD is in Northern Ireland"},
+            {"postcode": "IM4", "message": "The postcode IM4 is in the Isle of Man"},
+            {"postcode": "JE2 3FN", "message": "The postcode JE2 3FN is in Jersey"},
+            # English Postcode
+            {"postcode": "M2 3WQ", "message": "in order of closeness to"},
+            # Scottish Postcode
+            {"postcode": "AB11 5BN", "message": "These results cover England and Wales."},
+            # Lower case Postcode
+            {"postcode": "im4", "message": "The postcode IM4 is in the Isle of Man"},
+            # Invalid post code
```
HettieS, line 23: I'm not sure this test is actually for an invalid postcode, is that right? And the one below. It looks like from the message that they do a successful search. So the comment confuses me a bit. Could it be reworded to be clearer what sort of postcode input this is?

fala/apps/adviser/tests/test_crown_dependencies_view.py
```diff
@@ -8,46 +8,42 @@ class PostcodeValidationTest(SimpleTestCase):
...
+            {"postcode": "BT", "message": "in order of closeness to"},
+        ]
+
+        for case in test_cases:
+            with self.subTest(postcode=case["postcode"]):
+                data = {"postcode": case["postcode"]}
+                response = self.client.get(self.url, data)
+                self.assertContains(response, case["message"])
 
     def test_change_search_button_and_form_is_visible(self):
```
HettieS, line 35: test confused me a little - could we make what the test is doing a little more explicit in the method like:

def test_other_region_form_and_change_search_button_visible

fala/templates/macros/forms.html
```diff
@@ -54,9 +54,10 @@
       <div class="form-row {{ kwargs.row_class or '' }}">
         {% if field %}
           {% if field.errors and controlled_by and not controlled_by.errors or field.errors and not controlled_by %}
-            <input type="text" name="{{ field.name }}" value="{{ field.data or '' }}" class="govuk-input govuk-input--error govuk-!-width-one-third {{ field_attrs.class or '' }}" maxlength="{{ kwargs.max_length or '30' }}" id="id_{{ field.name }}">
+            <input type="text" name="{{ field.name }}" value="{{ field.data or '' }}" class="govuk-input govuk-input--error govuk-!-width-one-third {{ field_attrs.class or '' }}" maxlength="{{ kwargs.max_length or '30' }}" id="id_{{ field.name }}" aria-describedby="field-description-postcode">
```
HettieS, line 57: how come this branch of the condition doesn't need the field name check ie. `{% if field.name == 'postcode' %}aria-describedby="field-description-postcode"{% endif %}`? Unless I am totally misunderstanding what this branch does!

package-lock.json
```diff
@@ -7,7 +7,7 @@
     "@gulp-sourcemaps/map-sources": {
```
HettieS, line 7: what is this update? Should it be committed in this PR?

HettieS (commented):
Just a couple questions!

.gitignore
```diff
@@ -38,6 +38,7 @@ docs/_build
 env
```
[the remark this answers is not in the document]
psweeting1: `.venv` is another recommended way to create Python venvs. This wasn't in the ignore, much like the other envs you can see in the list above. 

It caught me out as I almost committed it. So putting it so stop other future developers.

fala/templates/macros/forms.html
psweeting1: We only want to apply it to the postcode input field as that is the only one that has help text. None of the others do. Because it's part of a loop, it actually applied the aria to every input.

.gitignore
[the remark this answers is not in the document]
HettieS: Ah okay, I meant more as it isn't related to the work - also sorry I should have been clearer, when I said 'file' I meant the .gitignore file in general rather than just the .venv bit. I normally try to only commit work in a PR that is related to the ticketed work, or I contain it in a separate commit or something, so it is clear

fala/templates/macros/forms.html
HettieS: Okay, maybe I just don't fully understand how this code is working! My thinking was you might also want the condition on line 57 ie `{% if field.name == 'postcode' %}aria-describedby="field-description-postcode"{% endif %}`

I thought this was the field with an error summary attached or something, and you might still want the aria description attached to it but I could be totally wrong!

fala/apps/adviser/tests/test_crown_dependencies_view.py
psweeting1: Your right, I've added in a comment explaining the "invalid postcode" which is technically valid situation. 

It's because we're not adding any numbers after the first two characters. 

I've removed BT as that proves nothing now looking at this again with your comments. 

So AB will search for Scottish addresses allowing us to pick up the region, it's a bad postcode though as it wont tell you which part of Scotland, e.g. AB1, AB2 etc. However we still pick up that it's a Scottish postcode. It's still not great but there is another ticket in the pipeline to improve postcodes and their regions as not all Scottish postcodes are Scottish postcodes. Fun times.

fala/apps/adviser/tests/test_crown_dependencies_view.py
psweeting1: I like it, changed.

fala/apps/adviser/tests/test_crown_dependencies_view.py
willc-work: these:
'BT' - does a search

'BT [and number] - goes to NI page.

'IM' - shows invalid postcode error message

'IM [and number] - goes to Isles of Man page.
Are all due to the regex being very simple and we should probably have a ticket to address that

psweeting1: I agree this ticket has highlighted the postcode issue even more.

fala/templates/macros/forms.html
psweeting1: Ohhhh no you're right. I've miss read this. I've added the if condition to the error summary to ensure it only adds the aria label if error is used.

.gitignore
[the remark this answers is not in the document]
psweeting1: I see you're point. I don't see the harm in adding this as it's not effecting the project and is just improving ways of working for developers. But I'm happy to take it out and put into a clean up PR which is needed. I think there is a ticket for that as the README.md is a mess.

HettieS (approved):

PR #290, opened 2024-06-24
DRY up category selection, moving methods

starswan:

While working on EL-1536, it was noticed that our view has too many parameters. Changing this was considered outside of scope for that ticket, so this PR makes that change independently. It then turned out the the code has duplication and confusing behaviour, so this has 2 things worth putting into the codebase 

This change doesn't have a specific ticket associated with it

## Checklist

### DIFF
--- a/fala/apps/adviser/templatetags/adviser_extras.py
+++ b/fala/apps/adviser/templatetags/adviser_extras.py
@@ -4,6 +4,7 @@ import re
 from urllib.parse import urlparse, parse_qs, quote
 from django_jinja import library
 from jinja2 import pass_context
+from laalaa.api import PROVIDER_CATEGORIES
 
 
 @library.filter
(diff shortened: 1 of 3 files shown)


psweeting1:
This is great! 

But I still see some of the code in the project that's not using the new helper functions, like this function

```
def _display_category(self):
            if "categories" in self._form.cleaned_data:
                categories = [PROVIDER_CATEGORIES[cat] for cat in self._form.cleaned_data["categories"]]
                formatted_categories = ", ".join(map(str, categories))

                return formatted_categories
            return []
```

MazOneTwoOne (approved):

MazOneTwoOne:
> This is great!
> 
> But I still see some of the code in the project that's not using the new helper functions, like this function
> 
> ```
> def _display_category(self):
>             if "categories" in self._form.cleaned_data:
>                 categories = [PROVIDER_CATEGORIES[cat] for cat in self._form.cleaned_data["categories"]]
>                 formatted_categories = ", ".join(map(str, categories))
> 
>                 return formatted_categories
>             return []
> ```

I think this is duplication and now covered by the `category_selection` method

PR #296, opened 2024-07-01
EL-1453 Replace gulp with django-webpack-loader

starswan:

Replace gulp with django-webpack-loader

Intentionally left blank.

## Checklist


fala/assets-src/sass/style.scss
```diff
@@ -1,7 +1,7 @@
-// Set up routing so requests for files in /static are served from /node_modules/govuk-frontend/dist/govuk/all.
 // import GDS components from `node_modules`
-$govuk-images-path: "/static/images/";
-$govuk-fonts-path: "/static/fonts/";
+//$govuk-images-path: "/assets/images/";
+//$govuk-fonts-path: "/assets/fonts/";
```
MazOneTwoOne, line 3: @starswan might want to get rid of these if not used.

fala/settings/base.py
```diff
@@ -90,6 +104,10 @@ def root(*x):
         "DIRS": [root("templates"), abspath(root(project_root, "node_modules", "mojular-templates"))],
```
MazOneTwoOne, line 104: @starswan we can rid of `mojular-templates`, I think

fala/settings/base.py
```diff
@@ -184,7 +202,13 @@ def root(*x):
 # Python dotted path to the WSGI application used by Django's runserver.
 WSGI_APPLICATION = "fala.wsgi.application"
 
-INSTALLED_APPS = ("django.contrib.sessions", "django.contrib.messages", "django.contrib.staticfiles", "requests")
+INSTALLED_APPS = (
```
MazOneTwoOne, line 205: @starswan I think this can be in square brackets

starswan: I think its more idiomatic for this to be a tuple - lists are mutable, this is not.

PR #303, opened 2024-07-05
Remove old files which were missed in previous tickets

starswan:

Several files have been found that should have been cleaned up from other tickets. This removes them to avoid polluting other change sets

## Checklist


MazOneTwoOne (approved):
Thank you for addressing the comments.

PR #308, opened 2024-07-10
EL-1481: Playwright setup

HettieS:

Implement playwright the Django way. All the tests live in the same place and are run via the Django admin stuff i.e. `./manage.py test`

This is because setting it up using the pytest plugin (as ticket suggested) didn't feel like the best way to do it. Getting pytest setup was making life harder and would mean either getting this to work for having two different ways to run tests, or pulling everything apart and getting it all set up with pytest. That didn't feel like the right step at this point, if we just want to start using playwright. 

I found having an extra playwright container was adding an extra level of complexity and creating differences between using `venv` and docker. The playwright tests can now just be run on the app container as you would run tests normally. There is no longer a playwright container. This means everything works the same for `venv` development, docker development and in the circle pipeline. I'm not sure if we might want a general test container at some point - but for now this feels more simple.

Also, the Dockerfile is now built in stages for base, dev, and prod so that the correct dependencies are included when you are working locally in Docker. This makes Docker development easier. 

Currently I have only added two test files, one for search page and one for results page. I think these can change/grow depending on the tests we decide to add. We can also consider moving methods into helpers.

I'm nervous about the production stage of the multi stage Docker image so please make sure this is correct/nothing has been missed.

## Checklist


psweeting1:
Looking really good! I think something that was on main related to the gulp removal has broken the tests in docker however. That code has been reverted. So I think a rebase will fix this. Once done, bash into the FALA container and run the `./manage.py test` again to check the tests are working okay going forward. I think you're good then and I can approve.

HettieS:
> Looking really good! I think something that was on main related to the gulp removal has broken the tests in docker however. That code has been reverted. So I think a rebase will fix this. Once done, bash into the FALA container and run the `./manage.py test` again to check the tests are working okay going forward. I think you're good then and I can approve.

Thanks @psweeting1 I've rebased, run docker and run the tests - all looking good now!

psweeting1 (approved):
LGTM

PR #319, opened 2024-07-18
EL-1559: Write FALA specific e2e playwright tests

MazOneTwoOne:

- used [Page Object Modelling](https://medium.com/analytics-vidhya/page-object-modeling-with-python-and-playwright-3cbf259eedd3) to represent our 3 main pages as code, to help us target specific parts of page, when writing the playwright tests. This means we keep to DRY principle. 

- have kept existing tests where there was good coverage, testing the view/display of pages. 
- have used playwright tests for the user journeys in FALA.

## Checklist


fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,72 @@
...
+        return self._page.locator("css=.govuk-error-summary")
+
+    def checkbox_by_label(self, label):
+        return self._page.get_by_label(label)
+
+    @property
+    def language_dropdown(self):
+        return self._page.locator('select[class="goog-te-combo"]')
+
+    def select_specific_page(self, page_no):
```
willc-work, line 67: this method also exists in SearchPage section of this file, feels like there is probably a way that this can be made to be accessible to all pages that need it.

fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def pagination_link_title(self):
+        return self._page.locator("span.govuk-pagination__link-title")
+
+    @property
+    def previous_link(self):
+        return self._page.locator('span.govuk-pagination__link-title:has-text(" Previous")')
+
+    def select_specific_page(self, page_no):
```
willc-work, line 26: this method also exists in SearchPage section of this file, feels like there is probably a way that this can be made to be accessible to all pages that need it.

fala/apps/adviser/tests/page_objects.py
MazOneTwoOne: @willc-work we can move these up into the `FalaPage` class , as the other page objects inherit from it. I've now made that change.  🙂

fala/apps/adviser/tests/page_objects.py
MazOneTwoOne: @willc-work we can move these up into the `FalaPage` class , as the other page objects inherit from it. I've now made that change.  🙂

fala/apps/adviser/tests/test_pagination_playwright.py
```diff
@@ -0,0 +1,19 @@
...
+from fala.playwright.setup import PlaywrightTestSetup
+
+
+class PaginationResults(PlaywrightTestSetup):
+
+    def test_results_pagination(self):
+        checkboxes = ["Crime"]
+
+        page = self.visit_results_page("M25 3JF", checkboxes)
+        expect(page.item_from_text("1576 results in order of closeness to M25 3JF.")).to_be_visible()
```
HettieS, line 11: I don't think we should hard code this number - I'd maybe omit that or possibly lose this line

fala/apps/adviser/tests/test_pagination_playwright.py
```diff
@@ -0,0 +1,19 @@
...
+        expect(page.pagination_link_title).to_be_visible()
+        page.next_link.click()
+        # this tests that changing page persists the category choice by not affecting the result count
+        expect(page.item_from_text("1576 results in order of closeness to M25 3JF.")).to_be_visible()
```
HettieS, line 15: oh I see why you have done this now with the number of results - still I'm not too sure this is the best way to test in case that number changes. Is it possible to target the first provider card and make sure the category `crime` is in the list for it?

HettieS: Because that's what we really want to know, rather than inferring it from the count

fala/apps/adviser/tests/test_results_page_journeys_playwright.py
```diff
@@ -2,20 +2,16 @@
...
-        self.page = self.browser.new_page()
-
-    def visit_valid_uk_results_page(self):
-        self.page.goto(f"{self.live_server_url}")
-        self.page.get_by_label("Postcode").fill("SE11")
-        self.page.get_by_role("button", name="Search").click()
-        expect(self.page.locator("h1")).to_have_text("Search results")
-
+class ResultPageEndToEndJourneys(PlaywrightTestSetup):
     def test_pagination(self):
```
HettieS, line 6: we might not need this test anymore as the paginator tests sort of cover this

fala/apps/adviser/tests/test_results_page_journeys_playwright.py
```diff
@@ -2,20 +2,16 @@
...
-        self.page.locator('span.govuk-pagination__link-title:has-text(" Next")').click()
-        expect(self.page.locator("h1")).to_have_text("Search results")
-        expect(self.page.locator('span.govuk-pagination__link-title:has-text(" Previous")')).to_be_visible()
+        page = self.visit_results_page("SE11")
+        expect(page.pagination_link_title).to_be_visible()
+        page.next_link.click()
+        expect(page.h1).to_have_text("Search results")
+        expect(page.previous_link).to_be_visible()
+
+    def test_change_search(self):
```
HettieS, line 13: we do already test this in the other region postcode tests but might be nice to check the persistence of the postcode here when they change search

fala/apps/adviser/tests/test_results_page_journeys_playwright.py
```diff
@@ -2,20 +2,16 @@
...
-    def setUp(self) -> None:
-        self.page = self.browser.new_page()
-
-    def visit_valid_uk_results_page(self):
-        self.page.goto(f"{self.live_server_url}")
-        self.page.get_by_label("Postcode").fill("SE11")
-        self.page.get_by_role("button", name="Search").click()
-        expect(self.page.locator("h1")).to_have_text("Search results")
-
+class ResultPageEndToEndJourneys(PlaywrightTestSetup):
```
HettieS, line 5: I wonder if it might be worth adding some tests here that check the UI displays as expected for the list of providers? Perhaps checking for the website links and google map links for one provider or something like that.

fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @property
+    def previous_link(self):
+        return self._page.locator('span.govuk-pagination__link-title:has-text(" Previous")')
+
+    @property
+    def change_search_grey_box(self):
+        return self._page.locator("li.govuk-body.notranslate")
+
+    def listitem_for(self, postcode):
```
HettieS, line 41: at the minute this just takes a postcode but you could also refactor to work for organisation and categories too.

fala/apps/adviser/tests/test_search_page_journeys_playwright.py
```diff
@@ -22,9 +14,25 @@ def test_postcode_search_journey(self):
...
             with self.subTest(postcode=postcode):
-                self.visit_search_page()
-                self.page.get_by_label("Postcode").fill(postcode)
-                self.page.get_by_role("button", name="Search").click()
-                expect(self.page.locator("h1")).to_have_text("Search results")
-                expect(self.page.get_by_role("listitem").filter(has_text=f"Postcode: {postcode}")).to_be_visible()
-                expect(self.page.get_by_text("in order of closeness")).to_be_visible()
+                page = self.visit_results_page(postcode)
+                expect(page.h1).to_have_text("Search results")
+                expect(page.listitem_for(postcode)).to_be_visible()
```
HettieS, line 19: this is doing the same thing as `page.change_search_grey_box.nth(1)).to_have_text("Organisation: Islington Law Centre"` (but with postcode). It would be nice to consolidate those.

fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def search_button(self):
+        return self.item_from_text("Search")
+
+    @property
+    def error_summary(self):
+        return self._page.locator("css=.govuk-error-summary")
+
+    @property
+    def no_results_alert(self):
```
HettieS, line 67: i think this lives in ResultsPage

fala/apps/adviser/tests/test_search_page_journeys_playwright.py
```diff
@@ -22,9 +14,25 @@ def test_postcode_search_journey(self):
...
+        )
+        expect(page.h1).to_have_text("Search results")
+        # this selector matches multiple things so picking out the 2nd one in the list.
+        expect(page.change_search_grey_box.nth(1)).to_have_text("Organisation: Islington Law Centre")
+
+    def test_no_results_journey(self):
+        page = self.visit_search_page()
+        page.organisation_input_field.fill("Test")
+        page.search_button.click()
+        expect(page.no_results_alert).to_be_visible()
```
HettieS, line 38: you could test for the content of the alert too if you want to make it extra regression proof

fala/apps/adviser/tests/test_crown_dependencies_playwright.py
```diff
@@ -0,0 +1,24 @@
+from playwright.sync_api import expect
+from fala.playwright.setup import PlaywrightTestSetup
+
+
+class CrownDependenciesTest(PlaywrightTestSetup):
```
HettieS, line 5: I would be quite keen to rename this spec 'OtherRegionsTest' because I think that's what this is?

fala/apps/adviser/tests/test_crown_dependencies_playwright.py
```diff
@@ -0,0 +1,24 @@
...
+    def test_jersey(self):
+        results_page = self.visit_results_page("JE1")
+        expect(results_page.h1).to_have_text("The postcode JE1 is in Jersey")
+        search_page = results_page.change_search()
```
HettieS, line 9: this is confusing. Based on the `def visit_results_page` I don't think they should be able to change search here, right? Or maybe it is a slightly different change button? I wonder if that could/should be clearer.

fala/apps/adviser/tests/test_crown_dependencies_playwright.py
```diff
@@ -0,0 +1,24 @@
...
+        # Turns out that Jersey postcodes don't preserve the search term
```
HettieS, line 10: yeah I think this must be by design - we don't want them using our service for Jersey. I would reword this comment to that affect. You could check for the absence of the Jersey postcode in the test maybe? So that it is intentional.

fala/playwright/setup.py
```diff
@@ -16,3 +18,59 @@ def tearDownClass(cls):
         super().tearDownClass()
         cls.browser.close()
         cls.playwright.stop()
+
+    def visit_results_page(self, postcode, checkbox_labels=None):
```
HettieS, line 22: these three methods for visiting the results pages are all very similar - could we create one method? Or two at most for the other regions if you think that is more readable.

fala/apps/adviser/tests/test_search_page_journeys_playwright.py
```diff
@@ -22,9 +14,25 @@ def test_postcode_search_journey(self):
...
-                expect(self.page.locator("h1")).to_have_text("Search results")
-                expect(self.page.get_by_role("listitem").filter(has_text=f"Postcode: {postcode}")).to_be_visible()
-                expect(self.page.get_by_text("in order of closeness")).to_be_visible()
+                page = self.visit_results_page(postcode)
+                expect(page.h1).to_have_text("Search results")
+                expect(page.listitem_for(postcode)).to_be_visible()
+                expect(page.item_from_text("in order of closeness")).to_be_visible()
+
+    def test_invalid_postcode_journey(self):
+        page = self.invalid_postcode_search()
```
HettieS, line 23: i would bring the lines from this invalid method directly into the test so it is readable. We aren't reusing this one and it is directly what we want to test.

fala/playwright/setup.py
```diff
@@ -16,3 +18,59 @@ def tearDownClass(cls):
...
+
+    def visit_search_page(self):
+        page = self.browser.new_page()
+        page.goto(f"{self.live_server_url}")
+        return SearchPage(page)
+
+    def invalid_postcode_search(self):
+        page = self.browser.new_page()
+        page.goto(f"{self.live_server_url}")
+        page.get_by_label("Postcode").fill("ZZZ1")
```
HettieS, line 65: i would be tempted to have the bulk of these methods in the actual test rather than the setup - isn't this the thing we are testing? We're kind of hiding it by having it here. Same for the invalid org search one.

package-lock.json
```diff
@@ -74,9 +74,9 @@
       }
```
HettieS, line 74: should this file be committed?

fala/playwright/setup.py
```diff
@@ -16,3 +18,59 @@ def tearDownClass(cls):
...
+        page.get_by_role("button", name="Search").click()
+        expect(page.locator("h1")).to_have_text("Find a legal aid adviser or family mediator")
+        return SearchPage(page)
+
+    def invalid_organisation_search(self):
```
HettieS, line 70: we're not actually using this anywhere as far as I can tell. I'd write this out as a test rather than having it in setup

fala/apps/adviser/tests/test_pagination_playwright.py
willc-work: It has to be hardcoded as we want to check that the number of results does not change when the pages are changed, as a change would indicate that the categorisation was not being persisted when the pagination was used. I agree that this number could change in practice as providers leave or join, but I think that will not be frequently and the assurance provided by hardcoding outweighs the work required to update this if the number of results changes.

fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,75 @@
...
+        return self._page.locator("h1")
+
+    @property
+    def language_dropdown(self):
+        return self._page.locator('select[class="goog-te-combo"]')
+
+    def item_from_text(self, text):
+        return self._page.get_by_text(text)
+
+    def select_specific_page(self, page_no):
```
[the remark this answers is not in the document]
willc-work: agree thats a better name

fala/apps/adviser/tests/test_pagination_playwright.py
willc-work: the count way makes it work across the entire search. We could check that a specific result on a page has the correct category, but we cannot safely extrapolate to all providers on that page. The count ensures that the basics of the search (category and postcode) do not change between pages.

fala/apps/adviser/migrations/0001_initial.py
```diff
@@ -5,7 +5,6 @@
 
 
 class Migration(migrations.Migration):
-
```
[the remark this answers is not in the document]
MazOneTwoOne: @HettieS this was added in when running `pre-commit run --all-files` - not something we did on purpose.

fala/apps/adviser/tests/page_objects.py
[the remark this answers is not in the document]
MazOneTwoOne: will change 🙂

fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,75 @@
...
+        return self._page.locator(f'a.govuk-pagination__link:has-text(" {page_no} ")')
+
+    def change_search(self):
```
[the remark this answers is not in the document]
MazOneTwoOne: @HettieS I think the reason we have the `change_search` method here is that we have a `change search` button on both the 'results page' and the 'other region page'. 

We want to return the SearchPage object as some tests want to use the methods in that object within test.

fala/apps/adviser/tests/page_objects.py
```diff
@@ -0,0 +1,75 @@
...
+        return self._page.get_by_role("listitem").filter(has_text=f"Postcode: {postcode}")
+
+
+class OtherRegionPage(FalaPage):
+    pass
+
+
+class SearchPage(FalaPage):
+    @property
+    def postcode_input_field(self):
```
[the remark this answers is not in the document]
MazOneTwoOne: Yes, have done this to make the tests more readable

fala/apps/adviser/tests/test_crown_dependencies_playwright.py
MazOneTwoOne: Thanks, I've update the comment to that effect:

`# We don't preserve Jersey postcodes search term, by design, as we don't want residents from Jersey to use our service`

fala/apps/adviser/tests/test_search_page_journeys_playwright.py
MazOneTwoOne: Agree, will move this across 🙂

package-lock.json
MazOneTwoOne: @willc-work did you mean to commit this file?

willc-work: think it was added in error, replaced it with the version from main

fala/apps/adviser/tests/page_objects.py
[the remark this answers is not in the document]
HettieS: Yeah I can see it is used in a couple places. I think it is probably just my (limited) understanding of Python making this look weird to me. I thought we were sort of forcing the output to be a search page (rather than letting the application render whatever comes after the click) but on reflection I guess the test would fail if something unexpected happened

fala/apps/adviser/tests/test_pagination_playwright.py
HettieS: okay i see what you mean - if there really is no other way to reliably check the categories have persisted then I guess this is fine at the min

fala/playwright/setup.py
MazOneTwoOne: Have kept to two for readability.

fala/apps/adviser/tests/page_objects.py
MazOneTwoOne: I've removed this, as it's not specific enough. We have an alert on Search & Results page.

fala/apps/adviser/tests/page_objects.py
MazOneTwoOne: `listitem_for` is specific for one test scenario, I don't see the value in making it more adaptable.  We have `visit_results_page` & `visit_results_page_with_full_search` which would cover majority of the journeys we want to test and it's very readable as to what you are testing when searching.


I could make it more adaptable with doing something like this below but I don't think it's worth it. 

> def listitem_for(self, text_type, text):
>     return self._page.get_by_role("listitem").filter(has_text=f"{text_type}: {text}")

fala/apps/adviser/tests/test_crown_dependencies_playwright.py
MazOneTwoOne: `change_search()` is looking for the "Change search" text. 

- There is a "Change search" button on ResultsPage with #changeSearchButton
- There is a "Change search" button on OtherRegionPage with #otherRegionChangeSearchButton

I think test is clear, but will add the above as a comment on the method.

fala/apps/adviser/tests/test_results_page_journeys_playwright.py
MazOneTwoOne: So I think we said that these static views were covered in other tests which included `test_search_view_function.py`. I have changed it's name to `test_fala_views_and_functions.py`

fala/apps/adviser/tests/test_results_page_journeys_playwright.py
MazOneTwoOne: Frustratingly, testing this scenario (persisting postcode) is hard to get right. I think it's a limitation of Playwright. This is the addition we would want in the test:

`expect(search_page.postcode_input_field).to_have_text("SE11")`

fala/apps/adviser/tests/test_search_page_journeys_playwright.py
MazOneTwoOne: Have made the test clearer to understand.

fala/playwright/setup.py
HettieS: let's revisit this too

HettieS (approved):

PR #323, opened 2024-07-24
EL-1619: Deploy staging with helm

HettieS:

Setting ground work for switching to deploying staging with helm. 

Intentionally left blank.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,6 +244,7 @@ workflows:
           type: approval
           requires:
             - build
+            - Legacy staging deploy
           filters:
             branches:
               only:
@@ -252,8 +253,6 @@ workflows:
       - deploy:
           name: Legacy staging deploy
           environment: staging
-          requires:
-            - staging_deploy_approval
           context:
             - laa-fala
             - laa-fala-live-staging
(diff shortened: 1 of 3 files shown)


helm_deploy/laa-fala/values/fala-staging.yaml
```diff
@@ -47,6 +47,8 @@ ingress:
   hosts:
     - host: laa-fala-staging-helm.cloud-platform.service.justice.gov.uk
```
willc-work, line 48: this should change to be:
laa-fala-staging.apps.live-1.cloud-platform.service.justice.gov.uk
mainly because there may be some users out there using that URL through a bookmark or something

.circleci/config.yml
```diff
@@ -290,7 +290,7 @@ workflows:
       - production_deploy_approval:
           type: approval
           requires:
-            - Legacy staging deploy
+            - Helm Staging Deploy
           filters:
```
willc-work, line 294: I think that `name: Legacy staging deploy` should have the `requires staging_deploy_approval` removed. We want that to deploy asap with the new URL and therefore release the old URLs. 

And staging_deploy_approval should have a `requires` that checks that `Legacy staging deploy` is complete, that way we know we can release the helm staging deploy

willc-work (approved):
LGTM

PR #327, opened 2024-07-29
EL-1607 Incorporate jinja2 template macros

starswan:

Convert FALA Html into macros from govuk-frontend-jinja 

Intentionally left blank.

## Checklist


fala/templates/adviser/search.html
```diff
@@ -35,68 +31,100 @@ <h1 class="govuk-heading-xl"> Find a legal aid adviser or family mediator </h1>
...
-                        >
-                        <label class="govuk-label govuk-checkboxes__label" for="id_categories_{{ loop.index0 }}_{{ label_text }}">
-                          {{ label_text }}
-                        </label>
-                      </div>
-                    {% endfor %}
-                  </div>
-                </fieldset>
-              </div>
+              {{ govukCheckboxes({
```
MazOneTwoOne, line 113: Can somehow get this added back in to the checkboxes `id="id_categories_{{ loop.index0 }}_{{ label_text }}`. It make the Google Analytics easier to understand as the Id for the categories selected will then have a name rather than just "category 2"

fala/templates/generic-py-base.html
```diff
@@ -41,29 +43,11 @@
...
-          </svg>
-        </a>
-        </div>
-        <div class="govuk-header__content">
-        <a href="/" class="govuk-header__link govuk-header__service-name">
-          Find a legal aid adviser or family mediator
-        </a>
-        </div>
-        </div>
+        {{ govukHeader({
```
MazOneTwoOne, line 46: I love how simple and easier this is to understand.

MazOneTwoOne (commented):
I noticed some small things which would be good to address and some L&D questions.

fala/apps/adviser/views.py
```diff
@@ -66,16 +74,67 @@ def get_queryset(self):
...
             pages = LaaLaaPaginator(self._data["count"], 10, 3, self._form.current_page)
+            current_page = pages.current_page()
             params = {
                 "postcode": self._form.cleaned_data["postcode"],
                 "name": self._form.cleaned_data["name"],
             }
+            categories = self._form.cleaned_data["categories"]
+
+            # create list of tuples which can be passed to urlencode for pagination links
+            category_tuples = [("categories", c) for c in categories]
```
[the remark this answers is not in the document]
starswan: It felt sensible to put it here - the research I found was that there is 'nothing special' about anything much in Django, so it seemed to make sense to me. I would normally try to keep logic out of templates in general, so this felt like a win

fala/templates/adviser/search.html
```diff
@@ -8,22 +8,18 @@
...
-            <ul class="govuk-list govuk-error-summary__list">
-              {{ Form.render_field_errors_for_summary(form.postcode) }}
-              {{ Form.render_field_errors_for_summary(form.name) }}
-            </ul>
-          </div>
-        </div>
+      {% if form.errors %}
+        {{ govukErrorSummary({
+          "titleText": "There is a problem",
+          "errorList": errorList
```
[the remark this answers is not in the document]
starswan: I've fixed the link - when the error source is _all_ then the target id doesn't work

fala/templates/adviser/search.html
starswan: Do we need the loop index? It's now back as it was but w/o the loop index e.g. id_categories_Crime

MazOneTwoOne: No the `loop index` just gave a number to the id, before the name was assigned to it. So it can be removed. Our analytics aren't dependant on it either.

fala/templates/adviser/search.html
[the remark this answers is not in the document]
MazOneTwoOne: This is a difficult one to reason about and will need Design input. I think your changes are better but as it's not stated on the ticket, it's hard to know if this is what others would also expect to happen when they see the error messaging.

Maybe highlight this new behaviour for the person UAT or get design to look as well.


Screenshot 2024-08-05 at 12 34 08

MazOneTwoOne (approved):
Approving with comment. Will need to make sure the Designers/Team are aware of new error messaging behaviour, when UAT'ing.

PR #331, opened 2024-08-13
EL-1690: Create Cookie Policy Page

MazOneTwoOne:

- create new cookies page
- add content
- update footer
- update base
- setup specific playwright setup in cookies test case, rather than using existing set-up
   - need to use `context` rather than `page`
- use `wait_for_selector`
- added in `wait_for_load_state()` because the page was not loaded fully, before test did it's thing.
- update wording of test and add a few more scenarios.

- n/a

## Checklist

### DIFF
--- a/fala/apps/adviser/tests/page_objects.py
+++ b/fala/apps/adviser/tests/page_objects.py
@@ -23,6 +23,15 @@ class FalaPage(object):
         self.item_from_text("Change Search").click()
         return SearchPage(self._page)
 
+    def accept_cookies_banner(self):
+        self._page.locator("#cookieAccept").click()
+
+    def reject_cookies_banner(self):
+        self._page.locator("#cookieReject").click()
+
+    def get_cookies(self):
+        return self._page.context.cookies()
+
 
 class ResultsPage(FalaPage):
     @property
@@ -69,3 +78,17 @@ class SearchPage(FalaPage):
     def search(self, postcode):
         self._page.locator("#id_postcode").fill(postcode)
         self._page.locator("#searchButton").click()
+
+
+class CookiesPage(FalaPage):
+    @property
+    def accept_cookies_radio_option(self):
+        return self._page.locator("#cookies-analytics")
+
+    @property
+    def reject_cookies_radio_option(self):
+        return self._page.locator("#cookies-analytics-2")
+
+    @property
+    def save_cookie_settings_button(self):
+        return self._page.locator("#saveCookies")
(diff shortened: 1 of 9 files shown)


fala/apps/adviser/tests/page_objects.py
```diff
@@ -69,3 +69,7 @@ def checkbox_by_label(self, label):
     def search(self, postcode):
         self._page.locator("#id_postcode").fill(postcode)
         self._page.locator("#searchButton").click()
+
+
+class CookiesPage(FalaPage):
+    pass
```
psweeting1, line 75: Pass? Don't think that needs to be there.

fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
```diff
@@ -0,0 +1,86 @@
...
+        page = self.visit_cookies_page_from_footer()
+        expect(page.h1).to_have_text("Cookies")
+
+    def test_cookies_page_view_accesed_from_banner(self):
+        page = self.visit_cookies_page_from_banner()
+        expect(page.h1).to_have_text("Cookies")
+
+
+class CookiesPageEndToEndJourneysWithFreshSetUp(StaticLiveServerTestCase):
+    def test_that_cookie_has_appropriate_value(self):
```
psweeting1, line 17: For now this is fine for this part of the cookie work. I'll add this to the next cookie banner work story. 

This is a really big test which can be broken up in re-useable functions. So what I'm thinking, this could be 
broken down into three tests that can be used as functions in a much bigger test.

- Accept cookie banner, cookie is made.
- Reject cookie banner, cookie is not added to session.
- Accept Cookie on cookie policy page. Cookie is made.
- Reject cookie on cookie policy page. Cookie is rejected.
- Now using the above tests as functions you can combine them all into a nice flow test. I accepted cookie in banner, go to cookie page, I reject cookie. You can then even have a test that flips the logic, reject then accept, etc. 

This provides better test coverage as you've have individual part of the components and it's journey.

fala/templates/generic-py-base.html
```diff
@@ -75,38 +79,31 @@
...
-            function(){
-              $("#cookieAccept").focus();
-            }
-          );
-          return false;
-        });
+      });
+      $("#cookie_policy_page_choice").submit(function(e){
+        e.preventDefault();
+        var value = $("input[name='cookies']:checked").val();
```
psweeting1, line 95: Again this fine as it's being replaced. But make sure in future to use `let` and not `var` here.

psweeting1 (approved):
Approving, but there needs to be more changes in the next cookie banner work ticket.

PR #335, opened 2024-08-16
EL-1655 FALA production helm deploy

starswan:

FALA production helm deploy

Circle CI changes appear to have already been made

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -305,6 +305,7 @@ workflows:
           environment: production
           requires:
             - production_deploy_approval
+            - production_deploy
           filters:
             branches:
               only:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -305,6 +305,7 @@ workflows:
           environment: production
           requires:
             - production_deploy_approval
+            - production_deploy
```
willc-work, line 308: added this step so that the helm deploy occurs after the kubernetes deploy has completed, this was we ensure that the ingress containing `find-legal-advice.justice.gov.uk` and `laa-fala-production.cloud-platform.service.justice.gov.uk` is deleted before we try to deploy with helm and create an ingress with the same values (which can result in a broken pipeline).

psweeting1 (approved):
LGTM

.circleci/config.yml
psweeting1: Great thanks!

PR #336, opened 2024-08-19
EL-1507: Pre-populated-url

MazOneTwoOne:

- form validation happens when the data is cleaned, i.e. form validation, form errors, form cleaned
- `def clean(self)` is the method triggering form validation, so have abstracted that into `AdviserSearchForm` form class

- have not made the form do a post, as suggested by the ticket, as this would not be feasible to do

## Checklist

### DIFF
--- a/fala/apps/adviser/tests/test_search_page_journeys_playwright.py
+++ b/fala/apps/adviser/tests/test_search_page_journeys_playwright.py
@@ -14,15 +14,18 @@ class SearchPageEndToEndJourneys(PlaywrightTestSetup):
         expect(page.checkbox_by_label("Housing Loss Prevention Advice Service")).to_be_checked()
 
     def test_landing_page_with_multiple_url_params(self):
+        hlpas = "Housing Loss Prevention Advice Service"
+        edu = "Education"
         page = self.visit_search_page_with_url_params("categories=hlpas&categories=edu")
         expect(page.h1).to_have_text("Find a legal aid adviser or family mediator")
         expect(page.error_list).not_to_be_visible()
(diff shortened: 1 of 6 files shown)


psweeting1:
When searching for more than one category, the page still returns errors.

Also it appears to break the category order in the checkbox area.

willc-work (approved):

fala/apps/adviser/tests/test_search_page_journeys_playwright.py
```diff
@@ -7,6 +7,23 @@ def test_landing_page(self):
...
+        expect(page.error_list).not_to_be_visible()
+        expect(page.checkbox_by_label("Housing Loss Prevention Advice Service")).to_be_checked()
+
+    def test_landing_page_with_multiple_url_params(self):
+        page = self.visit_search_page_with_url_params("categories=hlpas&categories=edu")
+        expect(page.h1).to_have_text("Find a legal aid adviser or family mediator")
+        expect(page.error_list).not_to_be_visible()
+        expect(page.checkbox_by_label("Housing Loss Prevention Advice Service")).to_be_checked()
+        expect(page.checkbox_by_label("Education")).to_be_checked()
+        page._page.reload()
```
psweeting1, line 22: Can we stick `Housing Loss Prevention Advice Service` into variables that we can use instead of having to type the same string out everytime.

Also a note as to why we refresh the page to perform the same checks.

MazOneTwoOne: @psweeting1 have added variables and added comment

PR #337, opened 2024-08-20
EL-1550: Update cookie banner

MazOneTwoOne:

- remove jquery
- remove modernizer as not used
- add cookies nunjuck banner
- add cookies JS to set it in the session
- update tests to make them more modular
- update circleci config to build assets for tests job
- curl to get nvm
- close & boot up,  to use nvm
- add in `nvm use`
- remove `.nvmrc` file before trying to install nvm

- n/a

## Checklist


fala/assets-src/scripts/cookies.js
```diff
@@ -0,0 +1,70 @@
+const cookieBanner = document.getElementById("cookie-banner");
+const defaultMessage = document.getElementById("default-message");
+const acceptedMessage = document.getElementById("accepted-message");
+const rejectedMessage = document.getElementById("rejected-message");
+
+// Set cookies
+function setCookie(name, value, days) {
```
psweeting1, line 7: Change to use arrow function.

`const setCookie = (name, value, day) => {`

fala/assets-src/scripts/cookies.js
```diff
@@ -0,0 +1,70 @@
...
+// Hide rejected message
+const hideCoookieRejecttMessage = document.getElementById("rejected-hide");
+if (hideCoookieRejecttMessage) {
+  hideCoookieRejecttMessage.addEventListener("click", () => {
+    rejectedMessage.hidden = true;
+    cookieBanner.hidden = true;
+  })
+};
+
+// Cookie poolicy form
```
psweeting1, line 55: poolicy

fala/templates/generic-py-base.html
```diff
@@ -14,36 +15,99 @@
...
-                </button>
-                <button id="cookieReject" class="govuk-button govuk-button--secondary" data-module="track-click" data-accept-cookies="false" data-track-category="cookieBanner" data-track-action="Cookie banner rejected">
-                  {% trans %}Reject analytics cookies{% endtrans %}
-                </button>
-                {% trans %}<a id="cookies_banner_view_link" class="govuk-link" href="/cookies">View cookies</a>{% endtrans %}
-              </div>
-            </div>
-          </div>
-        </div>
+      <div class="fala-cookie-banner">
```
psweeting1, line 18: Because this is so big (Which is correct) I think this is worth placing into it's own HTML file and then being included.

When naming a sub html file to be included prefix it with an underscore `_`, So for example `_CookieBanner.html`

fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
```diff
@@ -14,73 +14,147 @@ def test_cookies_page_view_accesed_from_banner(self):
...
-    def test_that_cookie_has_appropriate_value(self):
+    banner_text = "Cookies on Find a legal adviser or family mediator"
+
+    def check_no_cookies(self, my_cookies):
+        assert len(my_cookies) == 0, "Cookies were found, which is surprising"
+
+    def verify_cookie_value(self, expected_value, my_cookies):
+        cookie_found = False
+        for cookie in my_cookies:
+            print(f"Name: {cookie['name']}, Value: {cookie['value']}")
```
psweeting1, line 25: Hello print what brings you here. How strange, you're linter (pre-commit) should have caught that.

fala/assets-src/scripts/cookies.js
```diff
@@ -0,0 +1,70 @@
...
+// Cookie policy form
+const cookiePolicyForm = document.getElementById("cookie_policy_page_choice");
+if (cookiePolicyForm) {
+  cookiePolicyForm.addEventListener("submit", (e) => {
+    e.preventDefault();
+    const selectedValue = document.querySelector("input[name='cookies']:checked").value;
+    if (selectedValue === "Allowed") {
+      setCookie("cookiePermission", "Allowed", 30);
+      tagMan();
+      location.reload();
```
psweeting1, line 64: I don't think we should refresh the users browser in the cookie policy page. We don't do this for the banner so why do it for the cookie policy page? 
It's an unnecessary refresh and HTTP request when the user will do that themselves naturally when they navigate away, which is a useful HTTP request.

fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
psweeting1: In future if you feel there needs to be some type of logging for tests, I'd recommend using python logging and not print. Print is create for debugging.

MazOneTwoOne: @psweeting1 could you give me an example of how to implement that? There isn't any specific for python logging in the playwright docs. All the stack overflow examples, would use [print](https://stackoverflow.com/questions/77945378/cant-capture-browser-error-logs-in-playright-test-w-python) statements.

Do you mean [this](https://realpython.com/python-logging/)?

fala/assets-src/scripts/cookies.js
MazOneTwoOne: Actually @psweeting1 I can do this in the JavaScript. I'll remove reload from FALA.

fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
MazOneTwoOne: For the moment, I'll remove this.

psweeting1 (approved):
LGTM great job

PR #339, opened 2024-08-21
EL-1698 addressed LAALAA API being called for scotland postcode region

psweeting1:

Adds scotland region into form clean up function. Updated and refactored code and unit tests.

Try manually testing UAT branch for confidence. 

[UAT Link](https://el-1698-unsupporte-fala-staging.cloud-platform.service.justice.gov.uk)

## Checklist

- [specific postcode"](https://dsdmoj.atlassian.net/browse/EL-1698)

### DIFF
--- a/fala/apps/adviser/tests/test_other_regions_view.py
+++ b/fala/apps/adviser/tests/test_other_regions_view.py
@@ -9,19 +9,21 @@ class PostcodeValidationTest(SimpleTestCase):
 
     def test_region_postcodes(self):
         test_cases = [
-            {"postcode": "GY1", "message": "The postcode GY1 is in Guernsey"},
+            {"postcode": "GY1 2HU", "message": "The postcode GY1 2HU is in Guernsey"},
             {"postcode": "BT93 8AD", "message": "The postcode BT93 8AD is in Northern Ireland"},
-            {"postcode": "IM4", "message": "The postcode IM4 is in the Isle of Man"},
+            {"postcode": "IM4 2HT", "message": "The postcode IM4 2HT is in the Isle of Man"},
+            # Postcode with incorrect prefix, but still picks up that it's isle of man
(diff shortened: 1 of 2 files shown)


fala/apps/adviser/tests/test_other_regions_view.py
```diff
@@ -9,19 +9,21 @@ class PostcodeValidationTest(SimpleTestCase):
 
     def test_region_postcodes(self):
         test_cases = [
-            {"postcode": "GY1", "message": "The postcode GY1 is in Guernsey"},
+            {"postcode": "GY1 2HU", "message": "The postcode GY1 2HU is in Guernsey"},
             {"postcode": "BT93 8AD", "message": "The postcode BT93 8AD is in Northern Ireland"},
-            {"postcode": "IM4", "message": "The postcode IM4 is in the Isle of Man"},
+            {"postcode": "IM4 2HT", "message": "The postcode IM4 2HT is in the Isle of Man"},
+            # Postcode with incorrect prefix, but still picks up that it's isle of man
+            {"postcode": "IM4 TESTTTTTTTTTTTT", "message": "The postcode IM4 TESTTTTTTTTTTTT is in the Isle of Man"},
```
willc-work, line 16: My only issue is that IM4 TESTTTTTTTTTTTT is not in the isle of man and is in fact an invalid postcode. Whilst the outcode (first part) is valid the incode (2nd part) is not. I am not sure that we should be telling people that a postcode that does not exist, is somewhere. 

I think because the poor regex that was previously being used we are trying to keep that behaviour, even though I don't think that was ever added as an intentional feature, rather it was a bug due to a regex that was to broad.

I think though this is a design/content issue and the ticket is a bug. so I am happy to approve but I think we need a discussion as a team if this is actually good behaviour for the service.

willc-work (approved):

fala/apps/adviser/tests/test_other_regions_view.py
psweeting1: Created a ticket in regards to this.

PR #341, opened 2024-08-22
EL-1697 use branch hash in ECR tag name

starswan:

use branch hash in ECR tag name to prevent wrong tag being used during UAT. Doesn't work with production k8s deployment, but that is now unused

Intentionally left blank.

## Checklist

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -7,7 +7,7 @@ BRANCH_RELEASE_NAME=$(echo $CIRCLE_BRANCH | tr '[:upper:]' '[:lower:]' | sed 's:
 deploy_branch() {
 # Set the deployment host, this will add the prefix of the branch name e.g el-257-deploy-with-circleci or just main
   RELEASE_HOST="$BRANCH_RELEASE_NAME-fala-staging.cloud-platform.service.justice.gov.uk"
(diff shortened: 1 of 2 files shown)


bin/deploy.sh
```diff
@@ -17,7 +17,7 @@ deploy_branch() {
                 --namespace=${K8S_NAMESPACE} \
                 --values ./helm_deploy/laa-fala/values/fala-uat.yaml \
                 --set image.repository="$ECR_ENDPOINT" \
-                --set image.tag="$IMAGE_TAG" \
+                --set image.tag="branch-$CIRCLE_SHA1" \
```
willc-work, line 20: should this be: "$BRANCH_RELEASE_NAME-$CIRCLE_SHA1"?

Otherwise it will be looking for something like 
"branch-k1j3hrjbf2krjg2k3j" 
instead of 
"EL-1234-thisthing-32r487yrtiu2wy"

willc-work: Clarified with Stephen that this is fine, we tag evrything branch or main and then use a CIRCLE_SHA1, it is the CIRCLE_SHA1 that is unique and so we can match an image to a deploy. resolving this

willc-work (approved):

PR #346, opened 2024-08-30
EL-1716 remove certificate

willc-work:

removes the cert from the k8s deploy as it is not valid for the host name we use here

Intentionally left blank.

## Checklist

-[El-1716](https://dsdmoj.atlassian.net/browse/EL-1716)


psweeting1 (approved):
LGTM

PR #347, opened 2024-09-02
EL-1605: Upgrade Node to v20

HettieS:

Upgrade Node to a later version. Managed to get this working to v20 which is stable with LTS.

Had to make some amends to Gulp packages to get this working from 16 upwards.

- Had to make a couple amends to Dockerfile where I think the file structure changed with the new Node image and so we couldn't copy to the same location anymore in the container

## Checklist


willc-work (approved):
LGTM

PR #350, opened 2024-09-05
BUGFIX: update directory name to match namespace

willc-work:

Updates the directory name to match the namespace so we can deploy for different environments with the same circle job.

Intentionally left blank.

## Checklist

- [EL-1656](https://dsdmoj.atlassian.net/browse/EL-1656)


psweeting1 (approved):
Connected to EL-1656 LGTM

PR #355, opened 2024-09-23
EL-1709: Playwright screenshots on test fails

MazOneTwoOne:

- have a way to take screenshots for the playwright tests only
- amend typos
- initialise the page object
- use `self._page` for all the methods in this class, so they have the initialised page object
- update the set-up for FALA cookies specific tests
- circleci update

- n/a

## Checklist

### DIFF
--- a/fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
+++ b/fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
@@ -7,14 +7,24 @@ import datetime
 
 class CookiesPageEndToEndJourneys(PlaywrightTestSetup):
     def test_cookies_page_view_accessed_from_footer(self):
-        page = self.visit_cookies_page_from_footer()
-        expect(page.h1).to_have_text("Cookies")
-        self.test_failed_take_screenshot = False
+        try:
+            page = self.visit_cookies_page_from_footer()
+            expect(page.h1).to_have_text("Cookies")
+            self.test_failed_take_screenshot = False
+
+        except Exception as e:
+            self.take_screenshot()
+            raise e
 
     def test_cookies_page_view_accessed_from_banner(self):
-        page = self.visit_cookies_page_from_banner()
-        expect(page.h1).to_have_text("Cookies")
-        self.test_failed_take_screenshot = False
+        try:
+            page = self.visit_cookies_page_from_banner()
+            expect(page.h1).to_have_text("Cookies")
+            self.test_failed_take_screenshot = False
+
+        except Exception as e:
+            self.take_screenshot()
+            raise e
 
 
 class CookiesPageEndToEndJourneysWithFreshSetUp(StaticLiveServerTestCase):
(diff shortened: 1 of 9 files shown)


fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
```diff
@@ -31,8 +35,18 @@ def tear_down(self, page, context):
         page.context.close()
         context.browser.close()
 
+    def take_screenshot(self, page):
```
psweeting1, line 38: Can you import this from the playwright setup python file? It's the same code.

fala/apps/adviser/tests/test_other_regions_playwright.py
```diff
@@ -10,6 +10,8 @@ def test_jersey(self):
         # We don't preserve Jersey postcodes search term, by design, as we don't want residents from Jersey to use our service
         expect(search_page.item_from_text("Organisation name")).to_be_visible()
 
+        self.test_failed_take_screenshot = False
```
psweeting1, line 13: Okay there is a lot of repeat. I think it's worth looking into AfterEach with playwright. AfterEach should be called after each test is run. It can perform tasks like clean ups. in our case, we can check if the test failed or not to then take a screenshot.

MazOneTwoOne: `AfterEach` from what I could tell was a node.js specific thing and did not help with out python set-up.

psweeting1:
[docker compose version](https://playwright.dev/python/docs/writing-tests#using-fixtures) I would say look into After All or After Each with playwright. There is a lot of repeating in this PR. Some of the code is the same so we can just import it's usage where possible into others documents. 

it's just a case of setting up fixtures to perform tasks after a test is completed which I think needs to be looked into and applied across all tests.

fala/apps/adviser/tests/test_cookies_page_journeys_playwright.py
MazOneTwoOne: Not this particular method as we're not inheriting from this class.

We have a fresh set-up for cookies, as we want to keep the same context.

MazOneTwoOne:
> [using fixtures](https://playwright.dev/python/docs/writing-tests#using-fixtures) I would say look into After All or After Each with playwright. There is a lot of repeating in this PR. Some of the code is the same so we can just import it's usage where possible into others documents.
> 
> it's just a case of setting up fixtures to perform tasks after a test is completed which I think needs to be looked into and applied across all tests.

Thanks for the call rather discussing via github comments:

- To be able to use `fixtures` we need to setup `pytest`, which we aren't doing in FALA
- `AfterEach` is a node.js  thing and we can't do that here

Instead I've done the following to remove duplication and make it more pythonic:
- I've removed remove variable's to ascertain if test has failed at the `setup` and `teardown` level
- I used the `try`, `except`, `finally` pattern at the test level to make it more pythonic and be able to do something when we have an exception.
   - I tried to do something at the `setup` and `teardown` level, but there isn't a nice way to store the value of this, without creating even more variables at the test level to be use at the`setup` and `teardown` level (catch 22)

MazOneTwoOne:
> Agreed to make a spike into exploring playwright working separately away from Django framework.
> 
> An epic will be made in order to explore our testing approach with playwright in FALA and possibly future projects.

PR #356, opened 2024-09-25
EL-1751: `form_submission` event not firing in GA4

MazOneTwoOne:

- amend `id` for one of the questions
- remove `novalidate` in form
- adjust the JS and run only when cookies are allowed

- n/a

## Checklist

### DIFF
--- a/fala/assets-src/scripts/google_tag_manager_data_layer.js
+++ b/fala/assets-src/scripts/google_tag_manager_data_layer.js
@@ -1,0 +1,30 @@
+const fala_form = document.querySelector('#fala_questions');
+
+function getCookie(name) {
+  const value = `; ${document.cookie}`;
+  const parts = value.split(`; ${name}=`);
+  if (parts.length === 2) return parts.pop().split(';').shift();
+  return null;
+}
+
+const cookie_value = getCookie('cookiePermission') === 'Allowed';
+
+if (fala_form && cookie_value) {
+    fala_form.addEventListener('submit', function(e){
+
+    // variables for postcode and organisation name
+    const postcode = document.querySelector("#id_postcode").value;
+    const organisation = document.querySelector("#id_name").value;    
+
+    // variables for checkbox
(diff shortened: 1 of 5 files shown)


fala/templates/base.html
```diff
@@ -102,32 +101,38 @@ <h1 class="govuk-heading-xl govuk-!-margin-top-7">
...
-
-      window.dataLayer.push({
-        'event': "formSubmission",
-        'postcode': postcode,
-        'organisation': organisation,
-        'checkbox': selected,
-      });
-    });
-  };
+    function tagManDataLayer() { // Add Data Layer to Google Tag Manager
```
psweeting1, line 111: Looking at this, this could go into a JS doc as it isn't needed for inline purposes compared to the tagman init functions.

psweeting1: Can you take a look and see if you can get this to work within side an imported JS document instead of inline. We want to try and avoid inline JS as much as we can.

MazOneTwoOne: I did consider this, but my rationale for keeping this at the bottom, is that we also need to fire this when the 'cookies== allowed'.

I found it simpler to check value & debug at this this level than in the JS where we have to keep rebuilding it.

MazOneTwoOne: Moved into it's own JS file

fala/assets-src/scripts/google_tag_manager_data_layer.js
```diff
@@ -0,0 +1,30 @@
+const fala_form = document.querySelector('#fala_questions');
+
+function getCookie(name) {
+  const value = `; ${document.cookie}`;
+  const parts = value.split(`; ${name}=`);
+  if (parts.length === 2) return parts.pop().split(';').shift();
+  return null;
+}
+
+const cookie_value = getCookie('cookiePermission') === 'Allowed';
```
psweeting1, line 10: Does this course any page flashes of the cookie banner briefly when the DOM is rendering?

psweeting1 (approved):
LGTM

PR #357, opened 2024-09-26
EL-1692: `postcode.io` validates postcode and country

MazOneTwoOne:

- create a new function called `validate_postcode_and_return_country` which uses `postcode.io` to validate full & partial postcodes and to look up country.
- update `clean` method to use  `validate_postcode_and_return_country`
- add error logging to handle what happens when there is no validate country or when `postcode.io` is down. 
- add retries to the api call

- for Guernsey & Jersey the country comes back as 'Channel Islands', I've had to use `nhs_ha` key.
 
## Checklist


fala/apps/adviser/forms.py
```diff
@@ -69,67 +70,88 @@ def __init__(self, *args, **kwargs):
         kwargs.setdefault("label_suffix", "")
         super(AdviserRootForm, self).__init__(*args, **kwargs)
 
-    def clean(self):
+    def clean(self):  # noqa: max-complexity=13 (increasing the complexity of a function so flake8 is happy)
```
HettieS, line 73: this method `def clean` is handling a lot of logic, hence why Flake8 is also complaining. Rather than ignoring the Flake8 complaint, it would be better to try and split the logic out into small methods to make this more readable/so the method has one responsibility only

fala/apps/adviser/forms.py
```diff
@@ -69,67 +70,88 @@ def __init__(self, *args, **kwargs):
...
+            self._region = Region.ENGLAND_OR_WALES
+            return data
+
+        # Validate postcode if provided and then determine region
+        if postcode:
+            valid_postcode = self.validate_postcode_and_return_country(postcode)
+            if not valid_postcode:
                 self.add_error("postcode", _("Enter a valid postcode"))
+            else:
+                country, nhs_ha = valid_postcode
```
HettieS, line 95: Could you add a comment around the use of nhs_ha? It isn't clear from the code why we are using that so would be good to document the reasoning

fala/apps/adviser/forms.py
```diff
@@ -69,67 +70,88 @@ def __init__(self, *args, **kwargs):
...
-        elif re.search(r"^GY[9]", postcode_no_space, flags=re.IGNORECASE):
-            return Region.GUERNSEY
-        elif re.search(r"^GY[0-8]", postcode_no_space, flags=re.IGNORECASE):
-            return Region.GUERNSEY
-        elif postcode_no_space[:2] in SCOTTISH_PREFIXES:
-            return Region.SCOTLAND
-        else:
-            return Region.ENGLAND_OR_WALES
+        # If no region was detected, return None.
+        return getattr(self, "_region", None)
```
HettieS, line 116: should the region logic be here instead? This method is now just responsible for returning "None" which seems a little weird.

HettieS (commented):
A couple comments from me, be good to also get a python dev to review!

fala/apps/adviser/forms.py
MazOneTwoOne: Yes - now broken up a bit and no longer need this 🙂

fala/apps/adviser/regions.py
```diff
@@ -11,31 +11,3 @@
         "SCOTLAND",
```
psweeting1, line 11: Doesn't look like we're using Region anymore either.

fala/apps/adviser/forms.py
```diff
@@ -74,62 +78,99 @@ def clean(self):
...
+            else:
+                return False
 
-        except laalaa.LaaLaaError:
-            # If there is an exception from LAALAA for example a postcode is not supported by our data. e.g. IM4
-            # Tell the user there was an error, but don't stop the usage of the site.
-            self.add_error("postcode", "%s %s" % (_("Error looking up legal advisers."), _("Please try again later.")))
+        except requests.RequestException:
+            # If there is an exception from postcode.io, tell the user there was an error, but don't stop the usage of the site.
+            self.add_error("postcode", _("Error looking up legal advisers via our API. Please try again later."))
```
psweeting1, line 172: The old error message is fine. "Error looking up legal advisers" The API means nothing to the end user when they see it pop up on the screen.

Have you tested to see this error response works? If not, try purposely changing the URL to the postcode.io wrong and then calling the search function.

fala/apps/adviser/tests/test_other_regions_view.py
```diff
@@ -24,6 +22,7 @@ def test_region_postcodes(self):
             {"postcode": "im4", "message": "The postcode IM4 is in the Isle of Man"},
             # Invalid post code, results are found and search
             {"postcode": "AB11 9EE", "message": "Enter a valid postcode"},
+            {"postcode": "IM4 TESTTTTTTTTTTTT", "message": "Enter a valid postcode"},
```
psweeting1, line 25: Can you add a scotland postcode that actually is England into this test.

`TD15 1UY` and `TD15`

MazOneTwoOne: Done

fala/apps/adviser/forms.py
MazOneTwoOne: Restored the wording.

Yes, did test that the error response works.

- I changed it from this:

`POSTCODE_IO_URL = "https://api.postcodes.io/postcodes?q="`

- to this:

`POSTCODE_IO_URL = "https://api.postcode.io/postcodes?q="`

- and got this:

Screenshot 2024-10-02 at 09 50 44

psweeting1 (approved):
LGTM

PR #366, opened 2024-10-24
EL-1683: Use `govuk_frontend_jinja/template.html`

MazOneTwoOne:

- remove `external.html`
- rename to `generic-base.html`
- rename partial views
- update the base.html
- add extra html to footer
- add a `page_title` block
- take what was needed from jinja html template and move across
- rename files
- place html into more logical folders
- add in event listener and `defer` attribute to script
- update to "govuk-frontend": "5.7.1"
- remove redundant imports
- add `.vscode` preferences to gitignore
- remove `before_main_js` block as it's not used. - remove setting id of `google_translate_element` it's done on other pages and not needed on base.

Intentionally left blank.

## Checklist


fala/templates/generic_base.html
```diff
@@ -0,0 +1,68 @@
+{%- from "govuk_frontend_jinja/components/skip-link/macro.html" import govukSkipLink -%}
```
psweeting1, line 1: This file is not using GOVUK templates. You wouldn't have to make your own when you can just use one straight from GOVUK itself. 

I'd also rename the file to just `base.html`

psweeting1 (approved):
LGTM

PR #371, opened 2024-11-08
EL-1776 add security and robots endpoints

willc-work:

Adds 2 endpoints for NCSC to check but it is also good practice to have these
add security.txt from moj standard template
https://github.com/ministryofjustice/security-guidance/blob/3245a63a5e3816a17d253eadcf81b4f2ee97663f/docs/security.txt#L4
add robots to allow crawlers in prod only

Intentionally left blank.

## Checklist


fala/apps/adviser/views.py
```diff
@@ -4,11 +4,37 @@
...
 from django.urls import resolve, reverse
 from django.views.generic import TemplateView, ListView
+from django.http import HttpResponse
+import os
 
 from .forms import AdviserSearchForm, AdviserRootForm
 from .laa_laa_paginator import LaaLaaPaginator
 from .regions import Region
 
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
willc-work, line 14: should this be in settings.py?

fala/urls.py
```diff
@@ -2,7 +2,15 @@
...
 
 from django.urls import re_path
-from adviser.views import AccessibilityView, AdviserView, SearchView, PrivacyView, CookiesView
+from adviser.views import (
+    AccessibilityView,
+    AdviserView,
+    SearchView,
+    PrivacyView,
+    CookiesView,
+    robots_txt,
```
willc-work, line 11: i dont know if these need to match the style of the other views here but they feel like different pages so it doesnt seem too wrong but happy to amend if needed

fala/apps/adviser/views.py
psweeting1: Yes, this should be in settings / base.py 

So you can do something like this `settings.BASE_DIR` or whatever to access the BASE_DIR variable

fala/apps/adviser/views.py
```diff
@@ -4,11 +4,36 @@
...
+    if environment == "production":
+        content = "User-agent: *\nDisallow:\n"  # Allow all
+    else:
+        content = "User-agent: *\nDisallow: /\n"  # Disallow all for staging/UAT
+
+    return HttpResponse(content, content_type="text/plain")
+
+
+def security_txt(request):
+    file_path = os.path.join(BASE_DIR, "public", "security.txt")
```
psweeting1, line 29: import and link BASE_DIR to the settings variable.

fala/apps/adviser/views.py
```diff
@@ -4,11 +4,36 @@
...
+import os
 
 from .forms import AdviserSearchForm, AdviserRootForm
 from .laa_laa_paginator import LaaLaaPaginator
 from .regions import Region
 
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+def robots_txt(request):
```
psweeting1, line 17: I was thinking about consistency as the imports in the URL side doesn't match or other URL's

Up to you, but we could put it into a class to keep the consistency. Happy to discuss your thought on this.
```
class RobotsTxtView(View):
    def get(self, request):
        environment = os.getenv("ENVIRONMENT", "development").lower()

        if environment == "production":
            content = "User-agent: *\nDisallow:\n"  # Allow all
        else:
            content = "User-agent: *\nDisallow: /\n"  # Disallow all for staging/UAT

        return HttpResponse(content, content_type="text/plain")
```

fala/urls.py
psweeting1: This is fine, it looks odd as you're not importing a class but a function instead.

psweeting1 (approved):
LGTM

PR #377, opened 2024-11-13
EL-1532: Footer not expanding correctly

MazOneTwoOne:

- override style of template, so that body does not have `min-height`. This means that the footer can expand.

- n/a

## Checklist


psweeting1 (approved):
LGTM, let's raise a ticket to find where the JS injected styles are coming from?

psweeting1 (approved):
LGTM

PR #378, opened 2024-11-13
EL-1825: Remove FALA DB Part 2

MazOneTwoOne:

- remove `rds-postgresql-instance-output` from `envs.tpl`. We no longer had a DB so don't need these.

- to help deploy our code. 

## Checklist


willc-work:
there are also postgres references still in the values files e.g.:
postgresql:
  enabled: false
  can you also remove those as part of this PR

willc-work (approved):

PR #383, opened 2024-11-29
EL-1921: Add organisation name as a hidden text of the 'View on map' link

MazOneTwoOne:

- add hidden text of org name, for the 'View on map (opens in new tab)' links.

Intentionally left blank.

## Checklist

### DIFF
--- a/fala/templates/adviser/results.html
+++ b/fala/templates/adviser/results.html
@@ -109,6 +109,7 @@
                 <span class="notranslate" translate="no">{{ item.location.postcode }}</span>
             </div>
             <a class="govuk-link govuk-!-display-none-print" class="url" target="_blank" rel="noopener" href="https://www.google.com/maps/search/?{{ item|google_map_params|urlencode }}">
+              <p class="govuk-visually-hidden">{{ item.organisation.name }}</p>
               View on map (opens in new tab)
             </a>
             {% if item.categories|length %}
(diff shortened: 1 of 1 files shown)


fala/templates/adviser/results.html
```diff
@@ -109,6 +109,7 @@ <h2 class="govuk-body-l govuk-!-font-weight-bold govuk-!-margin-bottom-0 govuk-!
                 <span class="notranslate" translate="no">{{ item.location.postcode }}</span>
             </div>
             <a class="govuk-link govuk-!-display-none-print" class="url" target="_blank" rel="noopener" href="https://www.google.com/maps/search/?{{ item|google_map_params|urlencode }}">
+              <p class="govuk-visually-hidden">{{ item.organisation.name }}</p>
```
psweeting1, line 112: `<span class="govuk-visually-hidden">{{ item.organisation.name }}</span>`

Using a `<p>` within an <a> tag is semantically incorrect, as block-level elements are not allowed inside inline elements like <a>.

MazOneTwoOne: @psweeting1 I might be misunderstanding but in HTML5 you can wrap `<p>` elements in links/anchor tags - [I believe it is encouraged](https://html5doctor.com/block-level-links-in-html-5/
) from the searches I did.

psweeting1 (approved):
LGTM

PR #384, opened 2024-12-03
EL-1902: Handle LAALAA 404s

HettieS:

This PR does two things:
* raises a 404 error page when a user tries to access a page in the pagination that does not exist (via manipulating the URL)
* fixes two pagination issues: one cosmetic issue where we were displaying '1' at the bottom of the page when there was only one page, and a bigger issue where we were generating an additional, unnecessary page at the end of the pagination that created a 404 in LAALAA and subsequently a 500 in FALA (now would be 'Page not found' and we wouldn't see an error).

I targeted only handling 404s differently as I don't know enough yet about other errors and the data structure that might come from LAALAA. Error handling is done quite weirdly, and the form is weirdly linked into the data response so haven't touched that. Most sensible place to raise 404 felt like the view given the current code structure.

Previously, 404s were being passed down to the forms/view weirdly, which meant the view was receiving no data structure - so everything blew up at the point of pagination. Have made the movement of data more sensible and now catching the 'Invalid page' with the 404.

Needs some good UAT to make sure I haven't inadvertently broken how other errors are returned and to make sure pagination works as expected.

## Checklist

### DIFF
--- a/fala/apps/adviser/forms.py
+++ b/fala/apps/adviser/forms.py
@@ -184,7 +184,6 @@ class AdviserSearchForm(AdviserRootForm):
                 )
                 if "error" in data:
                     self.add_error("postcode", (data["error"]))
-                    return {}
                 return data
             except laalaa.LaaLaaError:
                 self.add_error(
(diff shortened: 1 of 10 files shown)


psweeting1:
Is it possible to do a test where we alter the page URL param to trigger the old 500 issue we were facing?

psweeting1 (approved):
LGTM

PR #388, opened 2024-12-11
EL-1982: Error Validation on Single-Category-Search Page

MazOneTwoOne:

- add new error state view for single search, so that it returns the correct template
- initialise the correct form & render correct view based on `tailored_results` being in URL
- add `region` property to `SingleCategorySearchForm` as it needs it to perform a search
- Create a BaseSearch class which is shared by AdviserSearchForm & SingleCategorySearchForm - to stop code duplication
- pass in  through contexts to `SingleSearchErrorState` view
- add in playwright tests 

- n/a

## Checklist


psweeting1 (approved):
LGTM

PR #392, opened 2024-12-17
EL-1984: Add category to results h1

HettieS:

For single category search, modify the H1 accordingly.

Intentionally left blank.

## Checklist

### DIFF
--- a/fala/apps/adviser/tests/test_fala_views_and_functions.py
+++ b/fala/apps/adviser/tests/test_fala_views_and_functions.py
@@ -237,7 +237,7 @@ class SingleCategoryResultsPageTest(SimpleTestCase):
     def test_single_category_results_for_hlpas(self):
         self.data = {"tailored_results": "true", "postcode": "SE11", "categories": ["hlpas"]}
         response = self.client.get(self.url, self.data)
-        self.assertContains(response, "Your closest legal aid advisers for Housing Loss Prevention Advice Service")
+        self.assertContains(response, "Your closest legal aid advisers for the Housing Loss Prevention Advice Service")
 
 
 class NewSearchViewTemplate(SimpleTestCase):
--- a/fala/apps/adviser/tests/test_fala_views_and_functions.py
+++ b/fala/apps/adviser/tests/test_fala_views_and_functions.py
@@ -220,6 +220,26 @@ class ResultsPageWithJustOrgTest(SimpleTestCase):
         self.assertEqual(content, "Organisation: fooLegal problem: Debt, Education   Change search")
 
 
+class SingleCategoryResultsPageTest(SimpleTestCase):
+    client = Client()
+    url = reverse("search")
+
+    def test_single_category_results_for_mental_health(self):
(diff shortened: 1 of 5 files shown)


fala/apps/adviser/tests/test_fala_views_and_functions.py
```diff
@@ -214,6 +214,26 @@ def test_search_parameters_box_contains_only_organisation_and_categories(self):
         self.assertEqual(content, "Organisation: fooLegal problem: Debt, Education   Change search")
 
 
+class SingleCategoryResultsPageTest(SimpleTestCase):
```
MazOneTwoOne, line 217: Question - should this not be in `fala/apps/adviser/tests/test_single_category_search_view.py` ?

HettieS: The reason I have put them in the results page tests is because I am testing the content on the results page, whereas the 'Search' page is the landing page, and those tests are checking the content that displays there, the redirecting etc

HettieS: And the code I have added is directly in the results template, so seemed reasonable

fala/templates/adviser/results.html
```diff
@@ -11,14 +11,26 @@
...
-  <div class="govuk-grid-column-full">
     {% if data.count < 1 %}
+    <div class="govuk-grid-column-full">
         <h1 class="govuk-heading-xl govuk-!-margin-bottom-5">No search results</h1>
         <p class="govuk-body-m" id="no-results-info">There are no results for your criteria.</p>
+    </div>
     {% else %}
-        <h1 class="govuk-heading-xl">Search results</h1>
+    <div class="govuk-grid-column-two-thirds">
+        {% if tailored_results %}
```
psweeting1, line 21: My thinking that this could grow in size and then we have this in a template that just get bigger and bigger.

What are you're thoughts in putting this into another template and importing it into the results.html. There are example of this already in this template.

fala/apps/adviser/tests/test_fala_views_and_functions.py
MazOneTwoOne: That make sense. Should we call the test/class `ResultsPageWithSingleCategorySearch` just so it's in keeping with the other naming of the tests? Ignore this if you think it's fine as is.

HettieS: the reason I didn't name it 'single category search' and just put 'single category' is because I feel that the naming convention could change seeing as we are going to support multiple categories as the feature. But these tests are for a specific single category. Does that make sense? I didn't really want to refer to 'single category search' as a thing. That was my logic. 

And then ResultsPageWithSingleCategory didn't feel quite right because the results do take a single category at the minute, where as this is something a little different. But I'm really splitting hairs. It is sort of a SingleCategoryResultsPage in my head.

fala/templates/adviser/results.html
HettieS: yeah you make a good point! Happy to pop this code into a template!

PR #402, opened 2025-01-09
BUGFIX: add requires for staging so deployment does not happen too soon

willc-work:

adds a requires to staging deployment so the image is built before we deploy

Intentionally left blank.

## Checklist


psweeting1 (approved):
LGTM

PR #403, opened 2025-01-13
EL-1992: File structure for CLA FALA search

MazOneTwoOne:

- moving footer views to common folder
- moving robots & security text files to common folder
- moving mixins to common folder
- moving constants to common folder
- move paginator to common folder
- move region and utils to common folder
- remove code duplication
- rename 'search' view to 'results' view, as it makes more sense
- add license to `cla_search` app
- move tests out of adviser folder
- update naming conventions

- decided to group view classes, as it would be easier to remember what each of these views relate to. 
- `forms.py`, template tags & tests in each individual app not touched on this thicket, as they are not explicitly called out to do in this ticket.  

## Checklist


fala/apps/cla_search/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
psweeting1, line 1: Thanks for putting a licence in. I think it is in the wrong part of the project though. Currently you've got this in cla_search, when it should just be in the main part of the project. 

You'll need to update the README as well to link to licenece. 

[https://github.com/ministryofjustice/template-repository?tab=readme-ov-file#setup-instructions](https://github.com/ministryofjustice/template-repository?tab=readme-ov-file#setup-instructions)

Basically just copy our MOJ template in the licence area.

MazOneTwoOne: I've moved the license, but not sure about linking it from the readme?

psweeting1 (approved):
LGTM

HettieS (approved):
This looks really good and is a great improvement to the old structure - looks much clearer. I had two thoughts:

- there is a line in the README about running the tests, I know this isn't the test ticket but some of that structure has changed, would you mind updating in case we forget? Here is the line, I think the path needs `adviser` replacing maybe: https://github.com/ministryofjustice/fala/blob/8a274cdd663c1e9b09b3b114c9690a551a1d4183/docs/virtual-env.md#L133
- I'm not 100% on `cla_search` as the naming just because it would be nice to focus on the function of our code in the naming, rather than coupling the name to another application (that our code shouldn't need to know anything about), but I don't have a great suggestion right now -  I do wonder about something like 'category_search' that is a bit more description of what the thing is doing

Other than that, looks good!

PR #409, opened 2025-01-15
El 2031 amend change search function

kmahern:

This pr removes the Change Search button, and adds a back link, to the other_region search results page.

It updates the tests to test for this. It also makes a small change to playwright setup, as I think the OtherRegionPage page_objects were not being loaded correctly.

Intentionally left blank.

## Checklist

### DIFF
--- a/fala/playwright/setup.py
+++ b/fala/playwright/setup.py
@@ -47,10 +47,7 @@ class PlaywrightTestSetup(StaticLiveServerTestCase):
         for label in checkbox_labels:
             page.get_by_label(label).check()
         page.get_by_role("button", name="Search").click()
-        if page.locator("#changeSearchButton").is_visible():
-            return ResultsPage(page)
-        else:
-            return OtherRegionPage(page)
+        return ResultsPage(page)
 
     def visit_results_page_with_full_search(self, postcode, organisation, checkbox_labels):
         page = self.browser.new_page()
@@ -60,10 +57,7 @@ class PlaywrightTestSetup(StaticLiveServerTestCase):
         for label in checkbox_labels:
             page.get_by_label(label).check()
         page.get_by_role("button", name="Search").click()
-        if page.locator("#changeSearchButton").is_visible():
-            return ResultsPage(page)
-        else:
-            return OtherRegionPage(page)
+        return ResultsPage(page)
 
     def visit_search_page_with_url_params(self, url_params):
         page = self.browser.new_page()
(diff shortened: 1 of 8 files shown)


fala/templates/adviser/other_region.html
```diff
@@ -5,18 +5,14 @@
 {% block content %}
   <div class="govuk-grid-column-full">
     <div id="google_translate_element" class="govuk-!-margin-bottom-6"></div>
+    <a href="{{ url('adviser') }}" class="govuk-back-link govuk-!-margin-bottom-5 govuk-!-display-none-print">
```
psweeting1, line 8: If we can, we try to avoid using pure HTML for GOVUK components and instead use GOVUK Jinja2 components. Whenever you can find a Nunjucks version via GOVUK GDS components, there will also be a Jinja2 one as well.

[Jinja2 componets](https://github.com/LandRegistry/govuk-frontend-jinja/tree/main/govuk_frontend_jinja/templates/components) For examples on how to use them, I often look at the Flask Skeleton as there is a whole section where we import and use all the Jinja2 components. They are nearly the same as Nunjucks, with some slight syntax differences (Because it's Python)

[GDS component](https://design-system.service.gov.uk/components/button/)

If you've seen or copied this from somewhere else in the project, that's a shame and ideally should be changed whenever we get a chance to because we try not to just use HTML.

fala/playwright/setup.py
```diff
@@ -47,7 +47,7 @@ def visit_results_page(self, **kwargs):
         for label in checkbox_labels:
             page.get_by_label(label).check()
         page.get_by_role("button", name="Search").click()
-        if page.locator("#changeSearchButton"):
+        if page.locator("#changeSearchButton").count() > 0:
```
psweeting1, line 50: I think you can use the following:

`if page.locator("#changeSearchButton").is_visible():` instead of count. Count does not feel right. Give it a try but I'm pretty checking visibility over counting is the preferred way to assert.

fala/templates/adviser/other_region.html
kmahern: 👍 Thanks for the info @psweeting1 , I have amended this to use the Use GOVUK Jinja2 component as suggested.

fala/playwright/setup.py
kmahern: 👍 amended to use is_visible() as suggested.

psweeting1 (approved):
LGTM

fala/templates/adviser/other_region.html
```diff
@@ -5,15 +5,28 @@
...
+
     <h1 class="govuk-heading-xl">The postcode {{ postcode }} is in {{ region }}</h1>
     <p class="govuk-body">This search only covers England and Wales.</p>
     <p class="govuk-body">
       Find out about <a class="govuk-link" href="{{ link }}">Legal Aid in {{ region }}</a>
       or try a different search.
     </p>
+
+    {% if not tailored_results %}
+      <form action="/" method="get" novalidate>
```
psweeting1, line 24: My question on this is just, do we need this to be in a form if we are clearing everything for the user?

Can the button be a Jinja2 button component?

Just a copy of another button I found in search.html that should help with the conversion. 
```
 {{ govukButton({
                'text': "Search",
                'type': "submit",
                'classes': "govuk-!-margin-bottom-2",
                'id': "searchButton",
                'name': "search"
              }) }}
```

kmahern: @psweeting1 👍 I have amended the button to use the Jinja2 button component. I have left it in a form for now as this was the existing behaviour and I didn't want to increase the scope of this ticket but maybe something we could consider in a future refactor?

MazOneTwoOne (approved):
LGTM

PR #412, opened 2025-01-16
EL-1972: Create FALA translation files

MazOneTwoOne:

- update jinja2 middleware
- update base settings
- update locale folder
- create `.po` file
- make `LOCALE_PATHS` a tuple
- update readme
- update translations

- this work does not need to be behind a feature flag
- on this PR, I'm setting variables in `cookies.html` & `privacy.html` to extract the text out of Jinja2 templates (it's necessary as the `makemessages` command in Django doesn't natively support Jinja2 templates). This is being highlighted as a duplication, when it shouldn't really be.

## Checklist


HettieS (approved):
This all looks good to me. It will be much easier to test this once we start enabling the translations. 

A couple things:
- would you mind explaining what `fuzzy` does? I've done a bit of reading around it but not fully sure I get it. Is the use of `fuzzy` the thing that negates us needing a feature flag? I also wonder if it might be good to put a comment explaining the use of `fuzzy` here
- The duplication error is not coming from the variables but seems to be instead coming from these 7 lines:

```
 <li>{{_('the pages you visit')}}</li>
          <li>{{_('the categories you use to search')}}</li>
          <li>{{_('the postcode you use to search')}}</li>     
          <li>{{_('how long you spend on each page')}}</li>
          <li>{{_('how you arrived at the service')}}</li>
          <li>{{_('what you click on while you visit the service')}}</li>
          <li>{{_('the device and browser you use')}}</li>
```

These are duplicated across Privacy and Cookies hence the failing check (I think). I think the code was already like this so not a massive issue but you could extract those out into a shared partial and see if that improves the duplication. 

Approving, but if you'd mind looking into those above two bits before merging!

MazOneTwoOne:
> This all looks good to me. It will be much easier to test this once we start enabling the translations.
> 
> A couple things:
> 
> * would you mind explaining what `fuzzy` does? I've done a bit of reading around it but not fully sure I get it. Is the use of `fuzzy` the thing that negates us needing a feature flag? I also wonder if it might be good to put a comment explaining the use of `fuzzy` here
> * The duplication error is not coming from the variables but seems to be instead coming from these 7 lines:
> 
> ```
>  <li>{{_('the pages you visit')}}</li>
>           <li>{{_('the categories you use to search')}}</li>
>           <li>{{_('the postcode you use to search')}}</li>     
>           <li>{{_('how long you spend on each page')}}</li>
>           <li>{{_('how you arrived at the service')}}</li>
>           <li>{{_('what you click on while you visit the service')}}</li>
>           <li>{{_('the device and browser you use')}}</li>
> ```
> 
> These are duplicated across Privacy and Cookies hence the failing check (I think). I think the code was already like this so not a massive issue but you could extract those out into a shared partial and see if that improves the duplication.
> 
> Approving, but if you'd mind looking into those above two bits before merging!


On the first point the use of `fuzzy` in `fala/locale/cy/LC_MESSAGES/django.po`, is an automatic thing done by the Django translator when running the `makemessages` command. It's the file saying that a human needs to come along and fill in the gaps. It's only at the top as it's just an artefact/'boiler plate' from when I ran the `makemessages` command.

I don't think there is any harm keeping it in. I think this is how it get's used in other projects.

```
#, fuzzy
# The English text changed from "Submit" to "Send".  Please check if "Cyflwyno" is still appropriate.
msgid "Send"
msgstr "Cyflwyno" 
```


On the second point, I'll look at extracting those bullet points into it's own partial. Great find. Thank you.

PR #416, opened 2025-01-22
EL-2044: Refactor tests

HettieS:

Rearranged test files and tests. We now have adviser, category search and common tests and utils arranged in the right places.

Broken down some of the files into smaller chunks and moved some stuff to `def setUp`

Added a helper file with a couple of convenience methods, that can be added to when test suite grows.

Intentionally left blank.

## Checklist


fala/apps/category_search/tests/test_search_view.py
```diff
@@ -5,7 +5,7 @@
 from django.conf import settings
 
 
-class SingleCategorySearchViewTests(TestCase):
+class SearchViewTests(TestCase):
```
MazOneTwoOne, line 8: Question 1: did you mean to leave the comment in at line 1?:

`# single_category_search.py`

Question 2: why is this test name singular and this name plural `fala/apps/adviser/tests/test_search_views.py`?

Question 3: Do we need to update the `docs/virtual-env.md`?

MazOneTwoOne (approved):
Happy to approve this. Must have been hard to determine what tests are shared/common and what are specific to `adviser` & `category_search`. There is definite overlap. 

I like the addition of the `test_utils`.

I just have some small questions.

fala/apps/category_search/tests/test_search_view.py
HettieS: ooh good spots! Let me correct those things you spotted :)

PR #429, opened 2025-02-11
El 1975 hook up translation link

kmahern:

Required.

Intentionally left blank.

## Checklist


willc-work (commented):
LGTM - feels more straight forward than I was expecting.

willc-work (approved):

fala/common/mixin_for_views.py
```diff
@@ -21,8 +21,24 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class TranslationMixin:
+    def translation_link(self, request):
+        user_language = request.COOKIES.get("django_language", "en")
+        if user_language == "cy":
+            return "<a class='govuk-body govuk-link' id='language_switcher_link'>English</a> / Cymraeg"
```
psweeting1, line 28: I've spotted a small problem. These anchor links do not have `href`. Because of that the link doesn't show a cursor icon when you hover over them. 

an <a> tag without an href attribute fails accessibility because it is not recognised as an interactive element by screen readers, keyboard navigation, and assistive technologies.

You will need to add `href="#"` However the anchor link may try and go to the URL, so below in the JS I have added some code to help prevent that.

psweeting1: You're also missing some other needed attributes 

These are all on DWP component so you can find anything you've missed and need there. 

    ```
  hreflang="cy"
      lang="cy"
      rel="alternate"
      aria-label="Newid yr iaith ir Gymraeg"
```

File: fala/templates/adviser/_language_switcher.html [hunk 2]
```diff
@@ -1,5 +1,10 @@
 <div class="govuk-grid-row govuk-!-display-none-print language-switcher">
-  <p class="govuk-body">
-    English / <a class="govuk-body govuk-link" href="/" target="_blank">Cymraeg</a>
-  </p>
+  <form id="language_switcher_form" action="{{ url('set_language') }}" method="post">
+    <input type="hidden" name="next" value="{{ request.get_full_path() }}">
```
psweeting1, line 3: Is this input needed?

fala/assets-src/scripts/translations.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Handles language switcher form submission when clicked.
+ */
+document.addEventListener("DOMContentLoaded", () => {
```
psweeting1, line 4: This code helps stop URL behaviours. 

```
/**
 * Handles language switcher form submission when the link is clicked or activated via keyboard.
 */
document.addEventListener("DOMContentLoaded", () => {
    const form = document.getElementById("language_switcher_form");
    const trigger = document.getElementById("language_switcher_link");

    if (form && trigger) {
        // Handle click event
        trigger.addEventListener("click", (event) => {
            event.preventDefault(); // Prevent default link behavior
            form.submit();
        });

        // Handle keyboard activation (Enter & Space keys)
        trigger.addEventListener("keydown", (event) => {
            if (event.key === "Enter" || event.key === " ") {
                event.preventDefault();
                form.submit();
            }
        });
    }
});
```

fala/common/mixin_for_views.py
kmahern: 👍 thanks @psweeting1 have made these updates. I think the form seems to work without the javascript to prevent the link behaviour but this is definitely worth testing.

fala/templates/adviser/_language_switcher.html
psweeting1: Many thanks!

PR #430, opened 2025-02-11
Feature/el 2102 allow hosts

psweeting1:

- Updated env.example with new ALLOW_HOSTS that work for our domains.
- Updated helm value to accept sub domains ending in cloud-platform
- Updated Docker contain
- Added missing Licence for project. (Not part of ticket, but needs adding in to align with MOJ standards.)

- Helm values, make sure you're happy with these changes and understand why we've changed them.

## Checklist

- [EL-2102](https://dsdmoj.atlassian.net/browse/EL-2102)

[EL-2102]: https://dsdmoj.atlassian.net/browse/EL-2102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Crown Copyright (Ministry of Justice)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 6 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
MazOneTwoOne, line 1: Nice that this is finally added in. Should we capture the change somewhere on the PR comments, as it's not part of the ticket

MazOneTwoOne (approved):
Approved. Just one comment, not a blocker.

LICENSE
psweeting1: I have already added to the PR comment? Unless I've miss understood what you mean by this sorry.

PR #434, opened 2025-02-24
EL-2117: Update layout for navigational & language links - old

MazOneTwoOne:

If applied, this change will update the layout for navigational and language links

- update to CSS
- 2 templates on for normal pages, one for other region
- re-do logic of feature flag, so it is easier to remove, when feature is released

- n/a

## Checklist

### DIFF
--- a/fala/templates/adviser/_language_and_navigation.html
+++ b/fala/templates/adviser/_language_and_navigation.html
@@ -1,0 +1,46 @@
+{% if FEATURE_FLAG_WELSH_TRANSLATION %}
+  <div class="govuk-grid-row language-and-back-container">
+    <div class="govuk-grid-column-one-half">
+      {% if tailored_results %}
+        {% if request.GET.getlist('categories')|length > 1 %}
+          {{ govukBackLink({
+            'classes': "back_link, govuk-!-display-none-print, govuk-!-margin-top-0",
+            'text': BACK_LABEL,
+            'href': "/check?categories="+category_code+"&sub-category="+sub_category_code,
+          }) }}
+        {% else %}
+          {{ govukBackLink({
+            'classes': "back_link, govuk-!-display-none-print, govuk-!-margin-top-0",
+            'text': BACK_LABEL,
+            'href': "/check?categories="+category_code,
+          }) }}
+        {% endif %}
+    {% endif %}
+    </div>
+    <div class="govuk-grid-column-one-half language-switcher govuk-!-display-none-print">
+      <p class="govuk-body">
+        English / <a class="govuk-body govuk-link" href="/" target="_blank">Cymraeg</a>
+      </p>
+    </div>
+  </div>
+{% else %}
+  <div class="govuk-grid-row language-and-back-container">
+    <div class="govuk-grid-column-one-half">
+      {% if tailored_results %}
+        {% if request.GET.getlist('categories')|length > 1 %}
+          {{ govukBackLink({
+            'classes': "back_link, govuk-!-display-none-print, govuk-!-margin-top-0",
+            'text': BACK_LABEL,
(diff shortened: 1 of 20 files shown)


fala/templates/adviser/_language_and_navigation.html
```diff
@@ -0,0 +1,26 @@
+<div class="govuk-grid-row language-and-back-container">
+  <div class="govuk-grid-column-one-half">
+    {% if tailored_results %}
+      {% if request.GET.getlist('categories')|length > 1 %}
+        {{ govukBackLink({
+          'classes': "back_link, govuk-!-display-none-print, govuk-!-margin-top-0",
```
willc-work, line 6: classes should be space separated not comma separated.

fala/templates/adviser/_language_and_navigation_other_region.html
```diff
@@ -0,0 +1,51 @@
...
+          {{ govukBackLink({
+            'classes': "back_link_other_region, govuk-!-margin-top-0",
+            'text': BACK_LABEL,
+            'href': "/check?categories="+category_code+"&sub-category="+sub_category_code,
+          }) }}
+      {% else %}
+        {{ govukBackLink({
+          'classes': "back_link_other_region, govuk-!-margin-top-0",
+          'text': BACK_LABEL,
+          'href': "/check?categories="+category_code,
```
willc-work, line 15: I am not familiar enough with concatenation in Django so had a look for how it is proposed to do it.
https://stackoverflow.com/a/4524851
According to that we should do something like:
`{% with query_params="categories="|add:category_code|add:"&sub-category="|add:sub_category_code %}
      {{ govukBackLink({
        'classes': "back_link govuk-!-display-none-print govuk-!-margin-top-0",
        'text': BACK_LABEL,
        'href': "/check?"|add:query_params|urlencode,
      }) }}
    {% endwith %}`
    
However I am not sure of this may be over engineering a solution and you may have considered and disregarded this. The primary reason I can see why this would be preferred over using `+` is that special characters and spaces will be properly escaped, however we know that our codes are all simple 2 or 3 letter strings so would be unaffected. It may cause issues if the category codes change or new ones are added but I consider that unlikely.

fala/templates/adviser/_language_and_navigation_other_region.html
```diff
@@ -0,0 +1,51 @@
+{% if FEATURE_FLAG_WELSH_TRANSLATION %}
+  <div class="govuk-grid-row language-and-back-container">
+    <div class="govuk-grid-column-one-half">
+      {% if tailored_results %}
+        {% if request.GET.getlist('categories')|length > 1 %}
+          {{ govukBackLink({
+            'classes': "back_link_other_region, govuk-!-margin-top-0",
```
willc-work, line 7: same issue with commas separating classes

willc-work (changes requested):
the changes requested are only the ones abut commas in classes as I am not sure of this aesthetic or if the incorrect formatting will break the styling.

The other one is really just to have a discussion

fala/templates/adviser/_language_and_navigation_other_region.html
MazOneTwoOne: @willc-work I think that is "over engineering" it as you say. 

The `category_code` & `sub_category_code` are based on what CLA send over, so I would imagine that would be locked down on their side. Also our `CategoryMixin` (which assign the values to `category_code` & `sub_category_code`) redirects if the the request doesn't match the strings inside `fala/common/category_manager.py`

willc-work: cool, i am happy with using "+" on the basis that what we get from CLA is known

MazOneTwoOne:
no longer required

PR #438, opened 2025-02-27
EL-2071: Handle `govuk-frontend pagination component`, in order to get ‘previous’ & ‘next’ text translations

MazOneTwoOne:

- if applied, this will add translation text for `previous` & `next` pagination links

- n/a

## Checklist

### DIFF
--- a/fala/locale/cy/LC_MESSAGES/django.po
+++ b/fala/locale/cy/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: fala\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-02-21 14:18+0000\n"
+"POT-Creation-Date: 2025-02-27 15:39+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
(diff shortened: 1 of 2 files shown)


fala/locale/cy/LC_MESSAGES/django.po
```diff
@@ -92,6 +92,14 @@ msgstr ""
 "oes gennych blentyn sydd wedi cael anaf i’r ymennydd yn ystod beichiogrwydd, "
 "genedigaeth neu yn ystod 8 wythnos gyntaf ei fywyd."
 
+#: fala/common/states.py:71
```
willc-work, line 95: does it matter that these numbers are out of order? e.g.  Is there any benefit while debugging/ trying to find messages where these ids are used and we would look for them? If not no worries, if they contribute in any meaningful way should these be moved to a position that matches the numerical value?

willc-work (approved):
One query about ids, not a blocker

fala/locale/cy/LC_MESSAGES/django.po
MazOneTwoOne: @willc-work So these paths & numbers refer to the document the text is in and the line which the text is placed on. i.e  for this `fala/common/states.py:71`, `states.py` file and line number 71.

In terms of order this is an automatic process, which lists them alphabetically (file path) then numerically (line within the file).

I think that is what you were asking. Sorry if I've misunderstood.

PR #452, opened 2025-03-27
EL-2201: Change ARIA label on Welsh link

MazOneTwoOne:

- If applied this removes the aria labels for language links, as per the recommendation from uservision

- [Uservision finding](https://uv3202-moj-fala.uservisionaccessibility.co.uk/welsh-toggle-link-not-pronounced-properly-by-screen-readers-observation.html)

## Checklist


willc-work (changes requested):
The ticket requires the aria-label to remain but for both of the labels to be in thew english language. I think this implements the uservision suggestion but not the ticket requirements

MazOneTwoOne:
> The ticket requires the aria-label to remain but for both of the labels to be in thew english language. I think this implements the uservision suggestion but not the ticket requirements

@willc-work have updated. :)

PR #454, opened 2025-03-27
EL-2200 Cookies back link

willc-work:
makes the link function likke a back buttton

Intentionally left blank.

## Checklist

### DIFF
--- a/fala/common/footer_views.py
+++ b/fala/common/footer_views.py
@@ -1,17 +1,26 @@
 # coding=utf-8
 from fala.common.mixin_for_views import CommonContextMixin, TranslationMixin
 from django.views.generic import TemplateView
+from django.utils.http import url_has_allowed_host_and_scheme
 
 
 class FooterView:
     def get(self, request, **kwargs):
         context = self.get_context_data(**kwargs)
+
+        referer = request.META.get("HTTP_REFERER", "/")
+        host = request.get_host()
+
+        previous_url = referer if url_has_allowed_host_and_scheme(referer, allowed_hosts={host}) else "/"
+
         context.update(
             {
+                "previous_url": previous_url,
                 "translation_link": self.translation_link(request),
                 "language": self.language(request),
             }
         )
+
         return self.render_to_response(context)
 
 
--- a/fala/common/footer_views.py
+++ b/fala/common/footer_views.py
@@ -7,20 +7,12 @@ from django.utils.http import url_has_allowed_host_and_scheme
 class FooterView:
     def get(self, request, **kwargs):
         context = self.get_context_data(**kwargs)
-
-        referer = request.META.get("HTTP_REFERER", "/")
-        host = request.get_host()
-
-        previous_url = referer if url_has_allowed_host_and_scheme(referer, allowed_hosts={host}) else "/"
-
         context.update(
             {
-                "previous_url": previous_url,
                 "translation_link": self.translation_link(request),
                 "language": self.language(request),
             }
         )
-
         return self.render_to_response(context)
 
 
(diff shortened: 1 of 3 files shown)


fala/common/footer_views.py
```diff
@@ -1,17 +1,26 @@
...
 from fala.common.mixin_for_views import CommonContextMixin, TranslationMixin
 from django.views.generic import TemplateView
+from django.utils.http import url_has_allowed_host_and_scheme
 
 
 class FooterView:
     def get(self, request, **kwargs):
         context = self.get_context_data(**kwargs)
+
+        referer = request.META.get("HTTP_REFERER", "/")
```
MazOneTwoOne, line 11: @willc-work I understand the rationale to add the `previous_url` to the `FooterView` so that it is across all the pages on the app. 

But as the `previous_url` is only used on the `CookiesView`, show we not give the context to that view only?

fala/templates/adviser/cookies.html
```diff
@@ -10,7 +10,17 @@
 
 {% set successTitle = _('Success') %}
 {% set bodyText = _('Your cookie settings were saved') %}
-{% set bodyHTML = _('<h3 class="govuk-notification-banner__heading">Your cookie settings were saved</h3><p>Other government services may set additional cookies. These will have a separate cookie policy and banner.</p><p class="govuk-body"><a class="govuk-notification-banner__link" href="/">Go back to the page you were looking at</a></p>') %}
+{% set bodyHTML %}
+  <h3 class="govuk-notification-banner__heading">
```
MazOneTwoOne, line 14: @willc-work I think you will have to run `python manage.py makemessages -l cy -i locale -i venv` in order to get the previous translations applied in the `django.po` file. As your moving the order of the text around Django translations will not pick it up. 

After you've aded in translation, you then have to run python `manage.py compilemessages -l cy` to get them applied locally.

MazOneTwoOne (changes requested):
Just some small change requests. 

As an aside: I had to run `pip install -r requirements/generated/requirements-dev.txt` locally in order to get the django utils `url_has_allowed_host_and_scheme` to be recognised. I must have not updated in a while (thank you) - I don't think that will affect deployment as that is run every time we build and deploy.

fala/templates/adviser/cookies.html
willc-work: thanks, i thought as the content was the same it would be ok but of course the line number have changed.

fala/common/footer_views.py
willc-work: Yes good catch, I initially thought I needed it on every page to be able to capture the referring page and then forgot to revisit that assumption when it worked.

fala/common/footer_views.py
```diff
@@ -25,3 +26,20 @@ class PrivacyView(FooterView, CommonContextMixin, TranslationMixin, TemplateView
 
 class CookiesView(FooterView, CommonContextMixin, TranslationMixin, TemplateView):
     template_name = "adviser/cookies.html"
+
```
MazOneTwoOne, line 29: You could also do this, where you are getting the context from `FooterView` so you just update the context, instead of overriding and building again. 

```
    def get(self, request, **kwargs):
        response = super().get(request, **kwargs)

        context = response.context_data

        referer = request.META.get("HTTP_REFERER", "/")
        host = request.get_host()
        previous_url = referer if url_has_allowed_host_and_scheme(referer, allowed_hosts={host}) else "/"

        context.update({"previous_url": previous_url})

        return self.render_to_response(context)
```

MazOneTwoOne (changes requested):

willc-work:
@MazOneTwoOne thank you for your help with this PR. It was much improved by your code review

PR #616, opened 2026-07-21
Update base image to secure version

willc-work:
As a result of Snyk identified vulns in the base image we use:
updates the base image and the associated commands to patch vulnerabilities shown here https://github.com/ministryofjustice/fala/actions/runs/29733699601/job/88324712772

clean snyk scan shown here https://github.com/ministryofjustice/fala/actions/runs/29830800162/job/88635636727?pr=616

Intentionally left blank.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,19 +14,26 @@ RUN ./node_modules/.bin/gulp build --production
 #################################################
 FROM python:3.15-rc-alpine AS base
 
-COPY --from=node_build home/node/fala/assets /home/app/fala/assets
-
 ENV LC_CTYPE=C.UTF-8
 
-# Runtime User
+# Create runtime user early, then use that user in child stages.
 RUN addgroup -g 1000 app && adduser -u 1000 -G app -D -h /home/app app
 
-# Install python and build dependencies
-RUN apk add --no-cache build-base pcre2-dev
+# Install OS dependencies required by Python builds and runtime scripts.
+RUN apk add --no-cache \
+  bash \
+  build-base \
+  gettext \
+  pcre2-dev
 
 ENV HOME=/home/app \
-  APP_HOME=/home/app
+  APP_HOME=/home/app \
+  PATH=/home/app/.local/bin:$PATH
 WORKDIR /home/app
+
+COPY --from=node_build --chown=app:app home/node/fala/assets /home/app/fala/assets
+
+USER app
 
 #################################################
 # DEVELOPMENT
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -38,21 +46,16 @@
 
 # Install Python dependencies for development, includes tests.
 COPY ./requirements/generated/requirements-dev.txt ./requirements.txt
-RUN pip3 install --user --requirement ./requirements.txt
-
-# Add .local/bin to PATH to allow playwright command execution
-ENV PATH=/home/app/.local/bin:$PATH
-RUN playwright install --with-deps
+RUN pip3 install --user --requirement ./requirements.txt \
```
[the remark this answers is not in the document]
willc-work: causes an issue as rc compiles without pinned packages, if i use a trixie I can pin the packages but i have vulnerabilities. i am dismissing this because i am accepting the trade off between safe code and knwn vulnerabilities.

Dockerfile
```diff
@@ -3,56 +3,69 @@ FROM docker.io/node:25.9.0-trixie@sha256:74ff139f927c4a233bf0757edefe1ee057d185d
...
+  && (playwright install --with-deps || playwright install)
 
 COPY fala/ fala/
```
joselopez-moj, line 61: This is not a blocker, I just wanted to flag that Playwright with Alpine is unsupported. So something to keep in mind so we can upgrade to Debian or any other valid alternative in near future before it breaks. [You can read about it here](https://playwright.dev/python/docs/docker#:~:text=Alpine%20Linux%20and%20other%20distributions%20that%20are%20based%20on%20the%20musl%20standard%20library%20are%20not%20supported.)

Dockerfile
```diff
@@ -3,56 +3,69 @@ FROM docker.io/node:25.9.0-trixie@sha256:74ff139f927c4a233bf0757edefe1ee057d185d
...
 
 RUN ./node_modules/.bin/gulp build --production
 
 #################################################
 # BASE IMAGE USED BY ALL STAGES
 #################################################
-FROM python:3.13-slim-trixie AS base
-
-COPY --from=node_build home/node/fala/assets /home/app/fala/assets
+FROM python:3.15-rc-alpine@sha256:c8a2b555f655e34e8616f94fbc08e4e54cc381a4a8437cce3e5736fd818f790c AS base
```
joselopez-moj, line 16: Raising a small concern around using `python 3.15` given it's way too new and its status is still `prerelease`, being 3.14 the LTS

